PR #11, opened 2016-03-10
Added count() method to utils.js.

### LINKED ISSUE #10: Add count() method to utils.
Method that returns the number of items from iterator.

szymonkups:
Fixes #10.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -176,6 +176,24 @@ const utils = {
 	},
 
 	/**
+	 * Returns the number of items return by the iterator.
+	 *
+	 *		utils.count( [ 1, 2, 3, 4, 5 ] ); // 5;
+	 *
+	 * @param {Iterable.<*>} iterator Any iterator.
+	 * @returns {Number} Number of items returned by that iterator.
+	 */
+	count( iterator ) {
+		let count = 0;
+
+		for ( let _ of iterator ) { // jshint ignore:line
+			count++;
+		}
+
+		return count;
+	},
+
+	/**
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -176,6 +176,24 @@ const utils = {
 	},
 
 	/**
+	 * Returns the number of items return by the iterator.
+	 *
+	 *		utils.count( [ 1, 2, 3, 4, 5 ] ); // 5;
+	 *
```
pjasiun: @memberOf utils.utils

Reinmar:
We already have such method in the test utils: https://github.com/ckeditor/ckeditor5-utils/blob/master/tests/_utils/utils.js#L21

szymonkups:
Fixed. It was used only for tests, then this method started to be needed also in development. I've removed `getIteratorCount` method from test utils, and moved it as `count()` to utilities.

Reinmar:
Please also remember about doing this change in the core (I think that no other repo uses this method).

tests/utils.js
```diff
@@ -6,9 +6,8 @@
 'use strict';
 
 import utils from '/ckeditor5/utils/utils.js';
-import utilsTestUtils from '/tests/utils/_utils/utils.js';
 
-const getIteratorCount = utilsTestUtils.getIteratorCount;
+const getIteratorCount = utils.count;
```
pjasiun: Wouldn't it be better to use: `const count = utils.count;`? `getIteratorCount` was there because of the name of the method and now it's messy.

PR #22, opened 2016-04-08
Implemented Keyboard util.

### LINKED ISSUE #20: Implement Keyboard util
With features like

* known key names object,
* parse keystroke pattern,
* KeystrokeInfo interface,
* getCode( key name / keystroke )

Reinmar:
Fixes #20.

### DIFF
--- a/src/keyboard.js
+++ b/src/keyboard.js
@@ -45,6 +45,12 @@ export function getCode( key ) {
 		keyCode = keyNames[ key.toLowerCase() ];
 
 		if ( !keyCode ) {
+			/**
+			 * Unknown key name. Only key names contained by the {@link utils.keyboard.keyNames} can be used.
+			 *
+			 * @errror keyboard-unknown-key
+			 * @param {String} key
+			 */
 			throw new CKEditorError( 'keyboard-unknown-key: Unknown key name.', { key } );
 		}
 	} else {
--- a/src/keyboard.js
+++ b/src/keyboard.js
@@ -1,0 +1,152 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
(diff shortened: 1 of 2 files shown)


src/keyboard.js
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * * `a-z`,
+ * * `0-9`,
+ * * `arrow(left|up|right|bottom)`,
+ * * `backspace`, `delete`, `enter`,
+ * * `ctrl`, `cmd`, `shift`, `alt`.
+ *
+ * @member {Object} utils.keyboard.knownKeyNames
+ */
+export const knownKeyNames = generateKnownKeyNames();
```
pjasiun: Is `known` necessary? We do not have variables with arrays of unknown codes ;)

Reinmar: Hm... It underlines that these are not all key names. But ok, I'll remove the word and improve the documentation.

Reinmar:
I've just recalled that the error docs are missing. I'll add them as well.

src/keyboard.js
```diff
@@ -0,0 +1,152 @@
...
+ * Converts a key name or a {@link utils.keyboard.KeystrokeInfo keystroke info} into a key code.
+ *
+ * Note: Key names are matched with {@link utils.keyboard.knownKeyNames} in a case-insensitive way.
+ *
+ * @method utils.keyboard.getCode
+ * @param {String|utils.keyboard.KeystrokeInfo} Key name (see {@link utils.keyboard.knownKeyNames})
+ * or a keystroke data object.
+ * @returns {Number} Key or keystroke code.
+ */
+export function getCode( key ) {
```
pjasiun: An example would be useful. It is not obvious what the format of the `key` is using only a docs.

Reinmar:
> I've just recalled that the error docs are missing. I'll add them as well.

Done.

src/keyboard.js
Reinmar: For now you've got tests. It's not the time to polish the docs.

PR #27, opened 2016-04-26
Implemented batchify().

### LINKED ISSUE #26: Implement batchify() – a function which converts diff() output into a applicable deltas
`diff()` outputs results like:

```
diff( 'aba', 'acca' ); // [ EQUAL, INSERT, INSERT, DELETE, EQUAL ]
```

Often, you may need to then generate a set of deltas which transform the left text (current) to the right one (future) or the opposite. However, the output from `batchify()` should be generic enough (if possible), so it can be used with JS strings or DOM text nodes in a similar way if needed.

First feature which requires this functionality is ckeditor/ckeditor5-typing. I'm not sure there will be more soon, but that's enough.

Reinmar:
Fixes #26.

### DIFF
--- a/src/batchify.js
+++ b/src/batchify.js
@@ -30,7 +30,7 @@
  * in order to transform it to the output.
  */
 export default function batchify( diff, output ) {
-	const batches = [];
+	const batch = [];
 	let left = 0;
 	let right = 0;
 	let lastOperation;
@@ -73,11 +73,11 @@ export default function batchify( diff, output ) {
 
 	pushLast();
 
-	return batches;
+	return batch;
 
 	function pushLast() {
 		if ( lastOperation ) {
-			batches.push( lastOperation );
+			batch.push( lastOperation );
 			lastOperation = null;
 		}
 	}
--- a/src/batchify.js
+++ b/src/batchify.js
@@ -6,8 +6,8 @@
 'use strict';
 
 /**
- * Creates set of operations which need to be applied to the input in order to transform
- * it to the output. Can be used with strings or arrays.
+ * Creates a set of operations which need to be applied to the input in order to transform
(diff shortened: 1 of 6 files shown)


src/batchify.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Creates set of operations which need to be applied to the input in order to transform
+ * it to the output. Can be used with strings or arrays.
+ *
+ *		const input = Array.from( 'abc' );
+ *		const output = Array.from( 'xaby' );
+ *		const batch = batchify( diff( input, output ), output );
+ *
+ *		batch.forEach( operation => {
```
pjasiun: I do not like the collision name with batch and operation. Especially because you will use it in the same place where you will use engine batch, so you will have 2 thinks called `batch`.

pjasiun: Also note that 'INSERT' and 'DELETE' in diff are called "actions" now.

Reinmar: In similar way you have model in the UI and model in the engine. The same name for things which are significantly different. 

If you're able to find a name which still makes sense, then I'm ok with it. But I don't want to change the name to some weirdo just to avoid collision with totally different package. Certain concepts may repeat and trying to find a new name every time you implement them is pointless.

scofalik: I don't like it either. Not only it collides with names we already have, but is not strictly connected with what it does. What I mean is that it could be called any way (i.e. `groupify`, `groupDiff`, `diffOnContext`), so there is no pressure on keeping this name. I am sure we will find a better name :)

scofalik:
I think that the best name for "things" in `diff` function would be just "differences". But I can live with that :P

Reinmar:
Is it ready to be merged?

PR #29, opened 2016-05-05
Extract utils methods to separated files

### LINKED ISSUE #12: Export each util separately
I should be able to do:
```
import { count } from '/ckeditor5/utils/utils.js';
```
Instead of:
```
import utils from '/ckeditor5/utils/utils.js';
const count = utils.count;
```

### LINKED ISSUE #21: Tests: utils.mapsEqual lacks 100% CC
Note: run only http://tests.ckeditor.dev:1030/#tests/is:unit,tag:module!utils

screen shot 2016-04-08 at 13 46 15

oskarwrobel:
Fixes #12 
Fixes #21 

Connected to:
- https://github.com/ckeditor/ckeditor5/pull/151
- https://github.com/ckeditor/ckeditor5-engine/pull/394
- https://github.com/ckeditor/ckeditor5-ui/pull/23


src/comparearrays.js
```diff
@@ -0,0 +1,46 @@
...
+ *		compareArrays( [ 0, 2 ], [ 0 ] );			// 'EXTENSION'
+ *		compareArrays( [ 0, 2 ], [ 1, 2 ] );		// 0
+ *		compareArrays( [ 0, 2 ], [ 0, 1 ] );		// 1
+ *
+ * @memberOf utils
+ * @param {Array} a Array that is compared.
+ * @param {Array} b Array to compare with.
+ * @returns {utils.ArrayRelation} How array `a` is related to `b`.
+ */
+export default function( a, b ) {
```
Reinmar: Let's better make them named functions. So `export default function fooBar() { ... }`. Like here: https://github.com/ckeditor/ckeditor5-utils/blob/master/src/dom/indexof.js#L15. We'll be consistent and it will be easier to debug.

oskarwrobel: ok, sure.

tests/collection.js
```diff
@@ -157,7 +150,7 @@ describe( 'Collection', () => {
 
 				collection.add( item );
 
-				expect( item ).to.have.property( 'id', '3' );
+				expect( item.id ).to.be.above( '3' );
```
Reinmar: I'm afraid that this will always be true, because since uid()'s return value is cached and increased every time the function is used, the latest id can be far more than 3.

I'll try to figure out what this test checks precisely and get back to you.

Reinmar: OK, so... hm, since we can't stub `uid()` it looks like the only way to check this is by adding e.g. 100 items (more than could be added above, in previous tests) with ids (like `collection.add( getItem( '0' ) )`) and then adding one without id and checking whether its 100 + 1.

Reinmar:
OK, I checked that the problematic test really checks the code, because when I changed `_getNextId()` to not look for collisions it failed.

Great job!

PR #36, opened 2016-05-20
Refactored configuration class.

### LINKED ISSUE #33: Config#define and #get are not very useful for feature developers
Moved from https://github.com/ckeditor/ckeditor5/issues/148

oskarwrobel:
Fixes #33 

Related to:
- https://github.com/ckeditor/ckeditor5-creator-classic/pull/4
- https://github.com/ckeditor/ckeditor5/pull/175

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -5,7 +5,6 @@
 
 'use strict';
 
-import isObject from './lib/lodash/isObject.js';
 import isPlainObject from './lib/lodash/isPlainObject.js';
 
 /**
@@ -21,17 +20,20 @@ export default class Config {
 	 * @param {Object} [configurations] The initial configurations to be set.
 	 */
 	constructor( configurations ) {
-		if ( configurations ) {
-			this._set( configurations );
-		}
-
 		/**
-		 * Store for whole configuration. Configuration is hide it private property to be not accessible
-		 * directly from Config instance to keep consistent API.
+		 * Store for whole configuration.
 		 *
 		 * @private
-		 * @member {ConfigSubset} utils.config#_config
+		 * @member {Object} utils.config#_config
 		 */
+		this._config = {};
+
+		/**
+		 * Set initial configuration.
+		 */
+		if ( configurations ) {
+			this._setObjectToTarget( this._config, configurations );
+		}
 	}
 
 	/**
@@ -64,27 +66,32 @@ export default class Config {
 	 *			color: 'red',
 	 *		} );
 	 *
-	 *		config.toolbar.collapsed; // true
-	 *		config.toolbar.color; // 'red'
+	 *		config.get( 'toolbar.collapsed' ); // true
+	 *		config.get( 'toolbar.color' ); // 'red'
 	 *
 	 * @param {String|Object} name The configuration name or an object from which take properties as
 	 * configuration entries. Configuration names are case-insensitive.
-	 * @param {*} [value=null] The configuration value. Used if a name is passed to nameOrConfigurations.
+	 * @param {*} value The configuration value. Used if a name is passed.
 	 */
 	set( name, value ) {
-		this._set( name, value );
+		this._setToTarget( this._config, name, value );
 	}
 
 	/**
 	 * Does exactly the same as {@link utils.Config#set} with one exception - passed configuration extends
 	 * existing one, but does not overwrite already defined values.
 	 *
+	 * This method is supposed to be called by plugin developers to setup plugin's configurations. It would be
+	 * rarely used for other needs.
+	 *
 	 * @param {String|Object} name The configuration name or an object from which take properties as
 	 * configuration entries. Configuration names are case-insensitive.
-	 * @param {*} [value=null] The configuration value. Used if a name is passed to nameOrConfigurations.
+	 * @param {*} value The configuration value. Used if a name is passed.
 	 */
 	define( name, value ) {
-		this._set( name, value, true );
+		const isDefine = true;
+
+		this._setToTarget( this._config, name, value, isDefine );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/config.js
```diff
@@ -66,17 +72,98 @@ export default class Config {
 	 * @param {*} [value=null] The configuration value. Used if a name is passed to nameOrConfigurations.
 	 */
 	set( name, value ) {
```
oskarwrobel: > That's all we need so far. Sounds like we need only this Config( userConfig ) and Config#define( [path?], defaults ).
@Reinmar I decided to keep `set` method because implementation looks very easy, but If you are sure that we don't need it I will remove it.

src/config.js
```diff
@@ -96,96 +183,45 @@ export default class Config {
...
+			this._set( key, configuration[ key ], isDefine );
+		}, this );
 	}
 }
 
-mix( Config, ObservableMixin );
+/**
+ * Helper class to recognize if current configuration is nested or the top.
+ */
+class ConfigSubset extends Config {}
```
oskarwrobel: @Reinmar What do you think about it?

src/config.js
```diff
@@ -66,17 +72,98 @@ export default class Config {
...
 	 */
 	set( name, value ) {
-		// Just pass the call to the original set() in case of an object. It'll deal with recursing through the
-		// object and calling set( name, value ) again for each property.
+		this._set( name, value );
+	}
+
+	/**
+	 * Does exactly the same as {@link utils.Config#set} with one exception - passed configuration extends
+	 * existing one, but does not overwrite already defined values.
```
Reinmar: I miss an explanation why is this method useful. What's its main use case.

src/config.js
```diff
@@ -66,17 +72,98 @@ export default class Config {
...
+	 *
+	 * @param {String|Object} name The configuration name or an object from which take properties as
+	 * configuration entries. Configuration names are case-insensitive.
+	 * @param {*} [value=null] The configuration value. Used if a name is passed to nameOrConfigurations.
```
Reinmar: Why `null`?

src/config.js
Reinmar: We can keep `set()`, but there's no use case for it now. Unless... the constructor is using it ;).

Reinmar: PS. `set()` docs are outdate. The code samples show incorrect access to the props.

src/config.js
```diff
@@ -66,17 +72,98 @@ export default class Config {
...
+	 *
+	 * @param {String} name The configuration name. Configuration names are case-insensitive.
+	 * @returns {*} The configuration value or `undefined` if the configuration entry was not found.
+	 */
+	get( name ) {
+		// The target for this configuration is, for now, this object.
+		let source = this;
+
+		// Whole configuration is stored in private property.
+		if ( !( source instanceof ConfigSubset ) ) {
```
Reinmar: This is weird. What's the point in checking this?

src/config.js
```diff
@@ -66,17 +72,98 @@ export default class Config {
...
+			// Configuration is empty.
+			if ( !source._config ) {
+				return;
+			}
+
+			source = source._config;
+		}
+
+		// The configuration name should be split into parts if it has dots. E.g. `resize.width` -> [`resize`, `width`].
+		const parts = name.toLowerCase().split( '.' );
```
Reinmar: That's actually the perfect reason to use `get()`. We can do `toLowerCase()`.

tests/config.js
```diff
@@ -95,64 +119,154 @@ describe( 'set', () => {
 			}
 		} );
 
-		expect( config ).to.have.property( 'resize' );
-		expect( config.resize ).to.have.property( 'minheight' ).to.equal( 400 );
-		expect( config.resize ).to.have.property( 'maxheight' ).to.equal( 800 );	// Not touched
-		expect( config.resize ).to.have.property( 'hidden' ).to.equal( true );
+		expect( config.get( 'resize' ) ).to.be.eql( {
```
Reinmar: We use `equal`. Beside, shouldn't it be `deep.equal()`?

Reinmar:
My general concern is that this implementation becomes more and more complicated. I may be wrong, but this functionality seems to be implemetable in a much simpler way. If you try to do it from scratch, I think that you should be able to figure out much simpler solution.

Reinmar:
However, before we proceed with this ticket any further, we should agree about https://github.com/ckeditor/ckeditor5-utils/issues/37, because we may need to add the global config feature back.

tests/config.js
oskarwrobel: I treat `eql` as short version of `deep.equal`
http://chaijs.com/api/bdd/#method_eql

Reinmar: Yy... That's a confusing shorthand. I wouldn't thought it's any different from `equal()`. Let's better use the full name.

oskarwrobel:
I've just pushed another refactoring. I think code might be a little bit simpler, but I think we should keep in mind that it is not trivial feature especially because of possibility to set object as value and deep name 
```
set( 'some.deep.name', {
  foo: {
    bar: {
      lorem: 'ipsum'
    }
  }
} );
```

I haven't replaced `config.some.value` with `config.get( 'some.value' )` in related code yet and a lot of test fails.

Reinmar:
I pushed two small commits. I've noticed just one thing that I don't like, but I'm not sure whether it changed in this branch or it was like this before. I don't like that object keys are lowercased on `set()`. See e.g. this test:

```js
		config.set( {
			resize: {
				minHeight: 400,		// Override
				hidden: true,		// Expand
				icon: {
					path: 'abc',	// Override
					url: true		// Expand
				}
			}
		} );

		expect( config.get( 'resize' ) ).to.be.deep.equal( {
			minheight: 400,		// Overridden
			maxheight: 800,		// The same
			hidden: true,		// Expanded
			icon: {
				path: 'abc',	// Overridden
				url: true		// Expanded
			}
		} );
```

This looks bad. Object with the same properties should be returned by `get()`. This means that we can (and should) normalize only in `set()`.

If that's an old behaviour, then please report this as an issue and assign to 0.1.0.

oskarwrobel:
> I've noticed just one thing that I don't like, but I'm not sure whether it changed in this branch or it was like this before.

It was like this before so I'll create a ticket for that.

Reinmar:
So R+ here.

Issue #43, opened 2016-06-14
Config class should not modify names of properties


oskarwrobel:
Right now every property added to the config is transformed to lowercase.
```javascript
config.set( {
    resize: {
        minHeight: 400
    }
} );

expect( config.get( 'resize' ) ).to.be.deep.equal( {
    minheight: 400
});
```

It could make some problems:
```
config.set( {
  data: {
    fooBar: 'biz'
  }
} );

const data = config.get( 'data' );
data.fooBar // undefined
```

Transformation to lower case should be done only for comparison inside `get()` method. Or maybe  we don't need this transformation at all?

Noticed by @Reinmar here: https://github.com/ckeditor/ckeditor5-utils/pull/36#issuecomment-225824494

pomek:
> Transformation to lower case should be done only for comparison inside get() method. Or maybe we don't need this transformation at all?

So, what is the decision?

Reinmar:
Since we have a clear naming scheme in CKEditor 5 (CKEditor 4's was super confusing), I guess we could totally remove the normalisation.

But I've just started thinking for whom the normalisation was added in the first place, and it's clear that for developers who integrate CKEditor with their websites. So the normalisation is needed on `set()` not `get()` what means that this ticket is incorrect...

However, if we keep normalising on `set()` we'll introduce situations like the one in the tests where magically object properties get changed. This is unacceptable. Imagine that you use CKE's config object to store some bigger object that your feature pass to e.g. an external library. That's going to break super fast.

So, to sum up – normalisation must be removed as it will break things.

This change can be done under this issue as it still matches the scheme.

PR #45, opened 2016-06-15
Config class does not transform keys in given configuration.

pomek:
Fixes: #43

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -67,7 +67,7 @@ export default class Config {
 	 *		config.get( 'toolbar.color' ); // 'red'
 	 *
 	 * @param {String|Object} name The configuration name or an object from which take properties as
-	 * configuration entries. Configuration names are case-insensitive.
+	 * configuration entries. Configuration names are case-sensitive.
 	 * @param {*} value The configuration value. Used if a name is passed.
 	 */
 	set( name, value ) {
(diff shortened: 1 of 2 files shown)


oskarwrobel:
Documentation should be updated.
- https://github.com/ckeditor/ckeditor5-utils/pull/45/files#diff-9d9a6cd82f41984872a66a3ab0d440c4L70
- https://github.com/ckeditor/ckeditor5-utils/pull/45/files#diff-9d9a6cd82f41984872a66a3ab0d440c4L103
- https://github.com/ckeditor/ckeditor5-utils/pull/45/files#diff-9d9a6cd82f41984872a66a3ab0d440c4L116

tests/config.js
```diff
@@ -157,6 +157,18 @@ describe( 'set', () => {
 		expect( config.get( 'date' ) ).to.be.an.instanceof( Date );
 		expect( config.get( 'instance' ) ).to.be.an.instanceof( SomeClass );
 	} );
+
+	it( 'should not transform keys to lowerCase', () => {
```
oskarwrobel: I think this test is not necessary. It could have been useful during development related issue, but keeping it might be confusing.

src/config.js
```diff
@@ -67,7 +67,7 @@ export default class Config {
 	 *		config.get( 'toolbar.color' ); // 'red'
 	 *
 	 * @param {String|Object} name The configuration name or an object from which take properties as
-	 * configuration entries. Configuration names are case-insensitive.
+	 * configuration entries. Configuration names are case-sensitive.
```
oskarwrobel: I'm not sure if information about case-sensitive is needed, because JavaScript is case-sensitive, but let us wait for @Reinmar.

Reinmar: We can keep it. This may be helpful for devs coming from other languages.

PR #56, opened 2016-08-04
t/50: Implement event delegation mechanism

### LINKED ISSUE #50: Implement event delegation mechanism between observables
Related to https://github.com/ckeditor/ckeditor5-ui/issues/35.

To avoid redundant code in the UI, `Observable` should support configurable event delegation mechanism.

oleq:
Closes #50.

### DIFF
--- a/src/observablemixin.js
+++ b/src/observablemixin.js
@@ -243,6 +243,43 @@ const ObservableMixin = {
 			boundObservables.clear();
 			boundAttributes.clear();
 		}
+	},
+
+	/**
+	 * Delegates selected events to another {@link ObservableMixin} instance.
+	 *
+	 *		A.pipe( 'eventA' ).to( B );
+	 *		A.pipe( 'eventA', 'eventB' ).to( C );
+	 *
+	 *		// eventA will also be fired by B and C along with someData.
+	 *		A.fire( 'eventA', someData );
+	 *
+	 *		// eventB will also be fired by C along with someData.
+	 *		A.fire( 'eventB', someData );
+	 *
+	 * @method utils.ObservableMixin#bind
+	 * @param {String...} bindAttrs Observable attributes that will be bound to another observable(s).
+	 * @returns {utils.BindChain}
+	 */
+	pipe( ...evtNames ) {
+		if ( !evtNames.length || !isStringArray( evtNames ) ) {
+			/**
+			 * All event names must be strings.
+			 *
+			 * @error observable-pipe-wrong-events
+			 */
+			throw new CKEditorError( 'observable-pipe-wrong-events: All event names must be strings.' );
+		}
+
+		return {
+			to: ( dest ) => {
+				for ( let name of evtNames ) {
+					dest.listenTo( this, name, ( ...args ) => {
+						dest.fire( name, ...args );
+					} );
+				}
+			}
+		};
 	}
 };
 
--- a/src/observablemixin.js
+++ b/src/observablemixin.js
@@ -246,23 +246,25 @@ const ObservableMixin = {
 	},
 
 	/**
-	 * Delegates selected events to another {@link ObservableMixin} instance.
+	 * Delegates selected events to another {@link ObservableMixin} instance. For instance:
 	 *
-	 *		A.pipe( 'eventA' ).to( B );
-	 *		A.pipe( 'eventA', 'eventB' ).to( C );
+	 *		observableA.pipe( 'eventX' ).to( observableB );
+	 *		observableA.pipe( 'eventX', 'eventY' ).to( observableC );
 	 *
-	 *		// eventA will also be fired by B and C along with someData.
-	 *		A.fire( 'eventA', someData );
+	 * then `eventX` is piped (fired by) `observableB` and `observableC` along with `customData`:
 	 *
-	 *		// eventB will also be fired by C along with someData.
-	 *		A.fire( 'eventB', someData );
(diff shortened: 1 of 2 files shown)


scofalik:
Is there a clear purpose why this is in `ObservableMixin`, not `EmitterMixin`? It looks like the added code do not use `ObservableMixin` characteristic things, and it could be useful also in other places in code where functionallity of `EmitterMixin` is enough.

scofalik:
I am also unsure whether `pipe` is a good name for this method, `delegate` or `delegateEvent` seems more clear and intuitive for me, though it is longer. If you aimed for shortest name possible, it is okay.

oleq:
> If there is not, I propose to move it to EmitterMixin. I would like to see it in EmitterMixin as it could be useful in other parts of code - which do not need stuff introcued by ObservableMixin.

A good point. I focused on UI and I didn't see the whole picture. In `EmitterMixin` a low–level implementation is possible, which would improve the performance IMO. I'll figure it out.

> I am also unsure whether pipe is a good name for this method, delegate or delegateEvent seems more clear and intuitive for me, though it is longer. If you aimed for shortest name possible, it is okay.

Yes, I wanted to keep it as short as possible but since it is already confusing I'll consider renaming the method 👍

oleq:
I'm wondering what is the right way to pass `EventInfo` and payload when delegating events, i.e. in case of delegation `emitterC#foo` -> `emitterB#foo` -> `emitterA#foo`, what should land in `emitterA#foo` event callback arguments?

```js
const emitterA = getEmitterInstance();
const emitterB = getEmitterInstance();
const emitterC = getEmitterInstance();

emitterC.delegate( 'foo' ).to( emitterB );
emitterB.delegate( 'foo' ).to( emitterA );

emitterA.on( 'foo', ( ...args ) => {
	// What should ...args look like?
	//
	// #1. Linear arguments:
	//
	// args = [ 
	//		{ EventInfo for emitterA#foo }, 
	//		{ EventInfo for emitterB#foo }, 
	//		{ EventInfo for emitterC#foo }, 
	//		dataFromC 
	// ] 
	//
	// #2. Recursive arguments:
	//
	// args = [ 
	//		{ EventInfo for emitterA#foo }, 
	//		[
	//			{ EventInfo for emitterB#foo }, 
	//			[
	//				{ EventInfo for emitterC#foo }, 
	//				dataFromC 
	//			]
	//		]
	// ] 
	//
	// #3. Something else?
} );

emitterC.fire( 'foo', dataFromC );
```

Usually what really matters for the listener is the real source of the event (`emitterC`) and `dataFromC` payload.

For example, in UI, if `emitterA` is `ListDropdownModel`, `emitterB` is `ListModel` and `emitterC` is `ListItemModel`, then a feature which uses `ListDropdown(Model)` wants to know which `ListItem(Model)` has been clicked and what data is passed along with the click, i.e. to apply corresponding style to the text like in Headings feature.

So IMO linear option makes more sense, because it's easier to reach both:
1. The real source of the event (second to the last arg) 
2. `dataFromC` payload, which usually has a lot of meaning for the listener (last arg). 

OTOH, I think that accessing stuff with braces `[]` and indexes is neither elegant nor comfortable. Maybe some other structure makes more sense?  Like extending `EventInfo` class to accommodate intermediate `EventInfos` inside `EventInfo for emitterA#foo`. Then, the structure could look like 

```js
args = [ 
	{ EventInfo for emitterA#foo }, 
	dataFromC 
] 
```

where

```js
{ EventInfo for emitterA#foo } = {
	...
	source: emitterA,	

	// Yes, bad naming, we could improve it.
	delegationChain: [ { EventInfo for emitterB#foo }, { EventInfo for emitterC#foo } ],
	delegationSource: emitterC
}
```

which shortens the following

```js
emitterA.on( 'foo', ( evt, dataFromC ) => {
	// Do sth with evt.delegationSource and dataFromC...
}
```

Or maybe it's OK to cheat the entire system and silently pass only `EventInfo for emitterC#foo`:

```js
emitterA.on( 'foo', ( evt, dataFromC ) => {
	// Do sth with evt.source == emitterC and dataFromC...
}
```

It brings a lot of magic because it's no longer clear what road the event followed from the real source (C) to A. But it's also the simplest way to do that.

WDYT?

scofalik:
Hard to say, because it was introduced for UI and you know what cases need to be solved or might appear. If we need intermediate data in them (idk, maybe for focusing, accessibility or setting some gfx), we need to implement it using 1) or 2) way. Maybe we can just go away with `source` and `delegationSource`? In any case I think that data should be available right away, so it can't be hidden somewhere deep in the structure.

So, my vote goes for
```
{ EventInfo for emitterA#foo } = {
    ...
    source: emitterA,   

    // Yes, bad naming, we could improve it.
    delegationChain: [ { EventInfo for emitterB#foo }, { EventInfo for emitterC#foo } ],
    delegationSource: emitterC
}
``` 
with, or even without `delegationChain`

scofalik:
I don't know how important it will be not to write different callbacks for delegated and non-delegated events.

Suppose, that you have:
```
emitterA.on( 'click', myCallback );
emitterB.delegate( 'click' ).to( emitterA );
```

We must make sure that no matter whether the event is fired by `emitterA` or `emitterB`, `myCallback` has to be processed correctly.

oleq:
As long as there's only one `EventInfo` object and `data` passed to the callback in either of the cases, we should be safe.

oleq:
New implementation in https://github.com/ckeditor/ckeditor5-utils/pull/59.

PR #57, opened 2016-08-05
Added: word priorities for EmitterMixin.

### LINKED ISSUE #55: EmitterMixin: introduce worded priority levels.
`EmitterMixin` should take constants as a way to describe callback priority. I propose 5 levels of priority:
* `highest`
* `high`
* `normal`
* `low`
* `lowest`

Callbacks with higher priority will be fired before callbacks with lower priority, so callback added with 'very_high' priority will be called before callback with 'normal' priority. When two callbacks have same priority level, they are fired in the order of adding.

I expect core features to mostly use `highest`, `normal` and `lowest`.  If needs be, we might introduce `very_high` and `very_low` priority. However right now most scripts do not use priority at all, so I belive that 5 levels will be enough for long time.

For debugging and testing purposes, an object with constants-to-numbers mapping will be exported, like discussed here: https://github.com/ckeditor/ckeditor5-utils/issues/8#issuecomment-236164721. I also thought about different solutions, but this one is simplest. My only worry is that people will use it also outside of tests, even if we write in documentation that they should not. I haven't thought how to solve this issue, though. I had some ideas but each of them is lacking. Still, I think it's important to have flexibility in priority setting for test purposes.

scofalik:
Fixes #55.

Related PR in `engine`: https://github.com/ckeditor/ckeditor5-engine/pull/552

That's all.

### DIFF
--- a/src/priorities.js
+++ b/src/priorities.js
@@ -1,0 +1,39 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * String representing a priority value.
+ *
+ * @typedef {'highest'|'high'|'normal'|'low'|'lowest'} utils.PriorityString
+ */
+
+/**
+ * Provides group of constants to use instead of hardcoding numeric priority values.
+ *
+ * @memberOf utils
+ */
+const priorities = {
+	/**
+	 * Converts a string with priority name to it's numeric value. If `Number` is given, it just returns it.
+	 *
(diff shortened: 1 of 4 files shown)


src/priorities.js
```diff
@@ -0,0 +1,39 @@
...
+	 */
+	get( priority ) {
+		if ( typeof priority != 'number' ) {
+			return this[ priority ] || this.normal;
+		} else {
+			return priority;
+		}
+	},
+
+	highest: 100000,
```
pjasiun: Why not `Infinity`?

pjasiun: Ok, I know why:
```
Infinity -1 < Infinity
> false
```

Issue #19, opened 2016-04-04
Tests: Create a util to compare HTML strings


Reinmar:
Part of a bigger task: https://github.com/ckeditor/ckeditor5-engine/issues/306

Such a tool is super popular in CKEditor 4 in which we need to operate on HTML strings in every single place of the editor.

In CKEditor 5 we'll be testing HTML strings less often, because (I presume) we'll have tools for comparing data model to some pattern, but still in some places we could use some HTML helpler. E.g. in https://github.com/ckeditor/ckeditor5-ui/issues/13.

pjasiun:
I really missed such tool testing renderer. It was even inconvenient to check the `tagName`, because browser may change it to upper case.

Reinmar:
> because browser may change it to upper case.

They always return uppercase ;)

pjasiun:
On the other hand such tool should check if there are no sibling text nodes. Using `outerHTML` was not enough because I found cases where renderer created every letter as a separate text node.

Reinmar:
To reduce the size of iteration 2, I'd move this ticket to iteration 3 which is about browser compatibility. So far we know that certain tests are failing due to poor HTML comparison. The renderer tests already exist and of course could use such a tool but that's not a critical thing at the moment. It will become critical once we want to introduce compatibility with more browsers.

Reinmar:
After merging this PR I realised that this function should not be made public. It should be a test util, not a function that we'll have to maintain in this or similar form forever (which is true if we made it public).

PR #67, opened 2016-08-30
Created util for normalizing HTML in tests

maxbarnas:
Fixes: https://github.com/ckeditor/ckeditor5-utils/issues/19.

### DIFF
--- a/tests/normalizehtml.js
+++ b/tests/normalizehtml.js
@@ -11,49 +11,49 @@ describe( 'utils', () => {
 			let actual = '<a style="border:1px;" class="" href="file://"></a>';
 			let expected = '<a class="" href="file://" style="border:1px;"></a>';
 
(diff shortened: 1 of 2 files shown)


tests/normalizehtml.js
```diff
@@ -0,0 +1,59 @@
...
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import normalizeHtml from '/ckeditor5/utils/normalizehtml.js';
+
+describe( 'utils', () => {
+	describe( 'normalizeHtml', () => {
+		it( 'should sort attributes', () => {
+			let actual = '<a style="border:1px;" class="" href="file://"></a>';
```
Reinmar: Use `const` for all `actual` and `expected`. I'll fix this.

PR #69, opened 2016-08-30
t/68: Provided an utility to translate keystroke strings depending on environment

### LINKED ISSUE #68: Provide an utility to translate keystroke strings depending on environment
A followup of https://github.com/ckeditor/ckeditor5-ui-default/issues/29.

For instance, a <kbd>CTRL+B</kbd> keystroke should be translated to <kbd>⌘B</kbd> on Mac to be displayed  correctly in a button tooltip and reflect the actual keystroke used for the feature (i.e. Bold).

oleq:
Closes #68.

### DIFF
--- a/src/keyboard.js
+++ b/src/keyboard.js
@@ -4,6 +4,7 @@
  */
 
 import CKEditorError from './ckeditorerror.js';
+import env from './env.js';
 
 /**
  * Set of utils related to keyboard support.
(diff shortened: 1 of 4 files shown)


src/keyboard.js
```diff
@@ -82,14 +83,34 @@ export function getCode( key ) {
...
+ *
+ * @method utils.keyboard.getEnvKeystrokeText
+ * @param {String} keystroke Keystroke text.
+ * @returns {String} Keystroke text specific for the environment.
+ */
+export function getEnvKeystrokeText( keystroke ) {
+	const split = splitKeystrokeText( keystroke );
+
+	if ( env.mac ) {
+		if ( split[ 0 ].toLowerCase() == 'ctrl' ) {
```
Reinmar: `SHIFT+CTRL+A` is not going to work. We need to either document that CTRL must always be first or handle this case. I guess that documenting a standard way how keystrokes should be defined will be best. This can be done in this module for simplicity – e.g. as a new type `KeystrokeText`.

Reinmar: I'd document that because I've just found this in our code:

```

		// Set the Ctrl+B keystroke.
		editor.keystrokes.set( 'CTRL+B', 'bold' );
```

And compare this with `parseKeystroke` documentation... :D

Reinmar: BTW. Shift and Alt should also be handled by this function.

BTW2. The standard for defining keystrokes in CKE4 is: 

screen shot 2016-09-06 at 16 37 50

Reinmar: OTOH, I've never liked CKE4's notation. I tend to write `Ctrl+Shift+L` which is more readable for me. Pitty, that @AnnaTomanek is on her leave.

Reinmar: BTW3. You can see on the screenshot that Ctrl doesn't necessarily go first... Although, I'd usually write it as a first modifier.

Reinmar: I can see that even alone I can trigger a longer discussion so I moved this problem to https://github.com/ckeditor/ckeditor5-utils/issues/74 to not block this PR.

PR #78, opened 2016-09-15
Merged EmitterMixin additional parameters into options object.

### LINKED ISSUE #76: Merge EmiterMixins#on's ctx and prior args into options object
Having them separately is a problem for us because we want to add another option there in the DomEmitterMixin object – the `useCapture`.

```js
on( evtName, callback, options )
```

oskarwrobel:
Fixes #76 

Related:
- https://github.com/ckeditor/ckeditor5-ui/issues/79
- https://github.com/ckeditor/ckeditor5-engine/issues/593
- https://github.com/ckeditor/ckeditor5-typing/issues/44

### DIFF
--- a/src/emittermixin.js
+++ b/src/emittermixin.js
@@ -34,19 +34,14 @@ const EmitterMixin = {
 	 * @method utils.EmitterMixin#on
 	 */
 	on( event, callback, options = {} ) {
-		if ( options.priority === undefined ) {
-			options.priority = 'normal';
-		}
-
 		createEventNamespace( this, event );
 		const lists = getCallbacksListsForNamespace( this, event );
-
-		options.priority = priorities.get( options.priority );
+		const priority = priorities.get( options.priority );
 
 		callback = {
 			callback: callback,
 			context: options.context || this,
-			priority: options.priority
+			priority: priority
 		};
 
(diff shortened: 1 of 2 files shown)


Reinmar:
Besides, I should review all 4 tickets at the same time, right?

oskarwrobel:
> Besides, I should review all 4 tickets at the same time, right?

Yes.

src/emittermixin.js
```diff
@@ -26,23 +26,27 @@ const EmitterMixin = {
...
 	 * the priority value the sooner the callback will be fired. Events having the same priority are called in the
 	 * order they were added.
-	 * @param {Object} [ctx] The object that represents `this` in the callback. Defaults to the object firing the
-	 * event.
+	 * @param {Object} [options.context] The object that represents `this` in the callback. Defaults to the object firing the event.
 	 * @method utils.EmitterMixin#on
 	 */
-	on( event, callback, priority = 'normal', ctx = null ) {
+	on( event, callback, options = {} ) {
+		if ( options.priority === undefined ) {
```
Reinmar: We must not modify objects that were passed to the methods.

Reinmar: Pity that we still can't use destructuring ;(

PR #80, opened 2016-09-16
t/79: Imported dom/toPix util to pixelize any value.

### LINKED ISSUE #79: Import pixelize function from BalloonPanelView to DOM utils
A followup https://github.com/ckeditor/ckeditor5-ui-default/pull/75/files/fa91bff4c9096c4bd1b70b9333c84fb995e6dd6f?diff=unified#r78917964.

The `pixelize` helper has to be re–used in the `StickyToolbarView` so it's a good time to move it to the utils.

oleq:
Closes #79.


Reinmar:
Sorry for not noticing this earlier, but could we make this util like this?

```js
const toPx = toUnit( 'px' );
const toEm = toUnit( 'em' );

toPx( 10 ); // 10px
toEm( 10 ); // 10em

toUnit( 'rem' )( 10 ); // 10rem
```

oleq:
Why, yes. A good point. I think it's unlikely to be used with units different than `px`, but I'm cool with it.

oleq:
Done!

src/dom/tounit.js
```diff
@@ -0,0 +1,24 @@
...
+ * @method utils.dom.toUnit
+ * @param {String} unit An unit like "px" or "em".
+ * @returns {utils.dom.toUnit.helper}
+ */
+export default function toUnit( unit ) {
+	/**
+	 * A function, which adds a pre–defined trailing `unit`
+	 * to the passed `value`.
+	 *
+	 * @function utils.dom.toUnit.helper
```
Reinmar: I wonder how this is going to look in API docs :D.

oleq: Hard to tell, really. But it makes more sense this way, IMO.

PR #82, opened 2016-09-21
Introduced Focus Tracker.

### LINKED ISSUE #81: Introduce Focus Manager
See: https://github.com/ckeditor/ckeditor5-link/issues/3

oskarwrobel:
Fixes #81 

Depends on: https://github.com/ckeditor/ckeditor5-ui/issues/76

### DIFF
--- a/src/focusmanager.js
+++ b/src/focusmanager.js
@@ -1,0 +1,123 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* global setTimeout, clearTimeout */
+
+import DOMEmitterMixin from '../ui/domemittermixin.js';
+import ObservableMixin from './observablemixin.js';
+import CKEditorError from './ckeditorerror.js';
+import mix from './mix.js';
+
+/**
+ * Observes if registered HTMLElements are focused. It is mainly used for keeping information if at least one of the editor component
+ * is currently focused which means that editor is focused.
+ *
+ * **Note** `focus` and `blur` listeners use event capturing, so it is only needed to register wrapper HTMLElement
+ * which contain other `focusable` elements. But note that this wrapper element has to be `focusable` too.
+ *
+ * @memberOf utils
+ * @mixes DOMEmitterMixin
+ * @mixes ObservableMixin
+ */
+export default class FocusManager {
+	constructor() {
+		/**
+		 * True when one of the registered elements is focused.
+		 *
+		 * @readonly
+		 * @observable
+		 * @member {Boolean} utils.FocusManager#isFocused
+		 */
+		this.set( 'isFocused', false );
+
+		/**
+		 * List of registered elements.
+		 *
+		 * @private
+		 * @member {Array<HTMLElement>} utils.FocusManager#_elements
+		 */
+		this._elements = [];
+
+		/**
+		 * Event loop timeout.
+		 *
+		 * @private
+		 * @member {Number} utils.FocusManager#_nextEventLoopTimeout
+		 */
+		this._nextEventLoopTimeout = null;
+
+		/**
+		 * Currently focused element.
+		 *
+		 * @private
+		 * @member {HTMLElement} utils.FocusManager#_focusedElement
+		 */
+		this._focusedElement = null;
+	}
+
+	/**
+	 * Adds specified element and and start listening `focus` and `blur` events on this element.
+	 *
+	 * @param {HTMLElement} element
+	 */
+	add( element ) {
+		if ( this._elements.indexOf( element ) >= 0 ) {
+			throw new CKEditorError( 'focusManager-add-element-already-exist' );
+		}
+
+		this.listenTo( element, 'focus', () => this._focus( element ), { useCapture: true } );
+		this.listenTo( element, 'blur', () => this._blur(), { useCapture: true } );
+		this._elements.push( element );
+	}
+
+	/**
+	 * Removes specified element and stop listening events on this element.
+	 * When element is focused then blur it firs.
+	 *
+	 * @param {HTMLElement} element
+	 */
+	remove( element ) {
+		if ( element === this._focusedElement ) {
+			this._blur( element );
+		}
+
+		const elementIndex = this._elements.indexOf( element );
+
+		if ( elementIndex > -1 ) {
+			this.stopListening( element );
+			this._elements.slice( elementIndex, 1 );
+		}
+	}
+
+	/**
+	 * Stores currently focused element and set {utils.FocusManager#isFocused} as `true`.
+	 *
+	 * @private
+	 * @param {HTMLElement} element Element which has been focused.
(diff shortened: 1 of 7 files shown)


src/focusmanager.js
```diff
@@ -0,0 +1,123 @@
...
+ * is currently focused which means that editor is focused.
+ *
+ * **Note** `focus` and `blur` listeners use event capturing, so it is only needed to register wrapper HTMLElement
+ * which contain other `focusable` elements. But note that this wrapper element has to be `focusable` too.
+ *
+ * @memberOf utils
+ * @mixes DOMEmitterMixin
+ * @mixes ObservableMixin
+ */
+export default class FocusManager {
```
oskarwrobel: I don't like this class name. I had a problem to stop thinking that `FocusManager` doesn't manage, but only listen to `focus` and `blur` and informs if at least one element is focused.

My ideas in f2f talk with @Reinmar was `FocusObserver` (but we already have such class in engine) or `ComponentFocusObserver`, but might be to complicated.

oskarwrobel: cc @pjasiun @scofalik

src/focusmanager.js
```diff
@@ -0,0 +1,123 @@
...
+		 * Currently focused element.
+		 *
+		 * @private
+		 * @member {HTMLElement} utils.FocusManager#_focusedElement
+		 */
+		this._focusedElement = null;
+	}
+
+	/**
+	 * Adds specified element and and start listening `focus` and `blur` events on this element.
```
oskarwrobel: Typo:
_**and and** start listening_

src/focusmanager.js
Reinmar: `FocusManager` is not the most fortunate name, but I can't find a better one... Maybe something like `FocusState` but I don't think that such odd names will be any better than the goold old `FocusManager`.

Reinmar:
WDYT about `FocusTracker`? Came to my mind while I've been writing the API docs.

Reinmar:
There's, unfortunately, a serious problem with the tests – they require browser to be focused while running them ;/ This is what we tried to avoid. I'm not sure if we succeeded so far, but I know from CKE4 that it's a big pain in the ass. Especially, when we start using a CI this may turn out to be troublesome.

After a brief talk with @pjasiun we agreed that the best thing that we can do is to:

1. Create a manual test.
2. In automated test, mock what we can mock.

oskarwrobel:
I've made refactoring and now tests should be browser independent.

Reinmar:
And what about a manual test?

oskarwrobel:
Done: [FocusTracker manual test](http://localhost:1030/build/modules/amd/tests/utils/manual/focustracker)

src/focustracker.js
```diff
@@ -0,0 +1,125 @@
...
+
+	/**
+	 * Clears currently focused element and set {utils.FocusTracker#isFocused} as `false`.
+	 * This method uses `setTimeout` to change order of fires `blur` and `focus` events.
+	 *
+	 * @private
+	 * @fires utils.FocusTracker#blur
+	 */
+	_blur() {
+		this._nextEventLoopTimeout = setTimeout( () => {
```
oleq: BTW: What problem does `setTimeout` solve here? I also see it in v4's [focusManager](https://github.com/cksource/ckeditor-dev/blob/t/14569/core/focusmanager.js#L170) but does it still make sense? It kinda flashes a red light, an async code out of nowhere TBH.

oleq: Hum. Now I see. It prevents flashing when the focus circles among elements belonging to the same "focus container". What if we implemented something like that instead:

```js
add( element ) {
   ...
   this.listenTo( element, 'blur', ( { propInEvtThatIForgot: theChildWhichHasBeenBlurred }  ) => {
      this._blur( theChildWhichHasBeenBlurred );
   }, { useCapture: true } );
}

_blur( theChildWhichHasBeenBlurred ) {
   if ( this._focusedElement.contains( theChildWhichHasBeenBlurred ) ) {
     // It's not a blur really. Just passing the focus between theChildWhichHasBeenBlurred and some other element belonging to this._focusedElement
     return;
   }
}
```

Do you think it is possible?

oskarwrobel: It is not only for elements belonging to the same container. It is to prevent flashing in general. E.g when focus is moved from one balloon panel to the other. I don't like this `setTimeout` too and I was thinking about dropping it, but after talk with @Reinmar we both agree that it is needed.

oleq: Alright. It gives me the creeps but let's stick to this solution.

oleq:
If minor the issues I pointed out are fixed, I think the overall implementation would be OK and we could go with it.

src/focustracker.js
```diff
@@ -0,0 +1,125 @@
...
+		this._focusedElement = null;
+	}
+
+	/**
+	 * Starts tracking the specified element.
+	 *
+	 * @param {HTMLElement} element
+	 */
+	add( element ) {
+		if ( this._elements.indexOf( element ) >= 0 ) {
```
oskarwrobel: I've started refactor it and I found that collection helps in some cases (removing element) but also makes code more complicated in other (checking if collection contains element). I think that collection might be a little to much for this case.

src/focustracker.js
```diff
@@ -0,0 +1,125 @@
...
+	 * Stores currently focused element and set {utils.FocusTracker#isFocused} as `true`.
+	 *
+	 * @private
+	 * @param {HTMLElement} element Element which has been focused.
+	 */
+	_focus( element ) {
+		clearTimeout( this._nextEventLoopTimeout );
+
+		this._focusedElement = element;
+		this.isFocused = true;
```
oskarwrobel: > Or maybe drop isFocused completely, exposing this.focusedElement as an only public API? Not a critical thing, anyway. Just thinking aloud.

I was thinking about dropping `isFocused`, but I decided that `isFocused` fits better as API property.

> Maybe some way to bind isFocused to _focusedElement? In fact, they are always in sync and tell the same story but in a slightly different way (always this.isFocused === !!this._focusedElement). Maybe

It's nice idea, but it will force to make `_focusedElement` observable, what is not really necessary.

src/focustracker.js
oleq: If not `Collection`, then [ES6 Set](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Set). Anything, really :)

src/focustracker.js
oleq: I was inspired by the [implementation](https://github.com/cksource/ckeditor-dev/blob/t/14569/core/focusmanager.js#L169-L174) in v4. On the second thought, it's unlikely to happen that `_blur()` will be called twice in a row since, unlike what is in v4, it's not a public API but a DOM listener callback. So maybe you're right after all ;-) This is why timeouts drive me crazy.

src/focustracker.js
oleq: OK. So let's KISS and leave it as it is.

PR #86, opened 2016-09-29
Added: dom.getCommonAncestor, dom.getParents.

### LINKED ISSUE #85: Introduce DOM helpers: getCommonAncestor, getParents
While doing changes in `ckeditor5-engine` I needed two methods that operate on DOM and `utils` repository is a nice fit for them.

`getCommonAncestor( nodeA, nodeB )` looks for a lowest common ancestor for two  given nodes.
`getParents( node )` returns all parents of given node and the node itself, starting from the root.

scofalik:
Fixes #85

### DIFF
--- a/src/dom/getancestors.js
+++ b/src/dom/getancestors.js
@@ -3,7 +3,7 @@
  * For licensing, see LICENSE.md.
  */
 
-/* globals Document */
+/* globals Node */
 
 /**
  * Returns all ancestors of given DOM node, starting from the top-most (root). Includes the given node itself. If the
@@ -17,8 +17,8 @@
 export default function getAncestors( node ) {
 	const nodes = [];
 
-	// Do not return Document object since it's not a `Node` and we are interested in `Node`s (and `DocumentFragment`s).
-	while ( node && !( node instanceof Document ) ) {
+	// We are interested in `Node`s `DocumentFragment`s only.
+	while ( node && node.nodeType != Node.DOCUMENT_NODE ) {
 		nodes.unshift( node );
 		node = node.parentNode;
 	}
--- a/src/dom/getancestors.js
+++ b/src/dom/getancestors.js
@@ -3,16 +3,22 @@
  * For licensing, see LICENSE.md.
  */
 
+/* globals Document */
+
 /**
- * Returns all ancestors of given DOM node, starting from the top-most (root). Includes the given node itself.
+ * Returns all ancestors of given DOM node, starting from the top-most (root). Includes the given node itself. If the
+ * node is a part of `DocumentFragment` that `DocumentFragment` will be returned. In contrary, if the node is
+ * appended to a `Document`, that `Document` will not be returned (algorithms operating on DOM tree care for `Document#documentElement`
+ * at most, which will be returned).
  *
- * @param {Element|Text} node DOM node.
- * @returns {Array.<Node>} Array of given `node` parents.
+ * @param {Node} node DOM node.
+ * @returns {Array.<Node|DocumentFragment>} Array of given `node` parents.
  */
 export default function getAncestors( node ) {
 	const nodes = [];
 
-	while ( node ) {
+	// Do not return Document object since it's not a `Node` and we are interested in `Node`s (and `DocumentFragment`s).
+	while ( node && !( node instanceof Document ) ) {
 		nodes.unshift( node );
 		node = node.parentNode;
 	}
--- a/src/dom/getancestors.js
+++ b/src/dom/getancestors.js
@@ -4,12 +4,12 @@
  */
 
 /**
- * Returns all parents of given DOM node, starting from the top-most (root). Includes the node itself.
+ * Returns all ancestors of given DOM node, starting from the top-most (root). Includes the given node itself.
  *
- * @param {Node} node DOM node.
+ * @param {Element|Text} node DOM node.
  * @returns {Array.<Node>} Array of given `node` parents.
  */
-export default function getParents( node ) {
+export default function getAncestors( node ) {
 	const nodes = [];
 
 	while ( node ) {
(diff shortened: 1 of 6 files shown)


scofalik:
Done.

src/dom/getancestors.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * Returns all ancestors of given DOM node, starting from the top-most (root). Includes the given node itself.
+ *
+ * @param {Element|Text} node DOM node.
+ * @returns {Array.<Node>} Array of given `node` parents.
```
Reinmar: Doesn't it return elements only? Or can it also be a docfrag? In any case, the documentation could be more precise.

src/dom/getancestors.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * Returns all ancestors of given DOM node, starting from the top-most (root). Includes the given node itself.
+ *
+ * @param {Element|Text} node DOM node.
```
Reinmar: `HTMLElement`

src/dom/getcommonancestor.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import getAncestors from './getancestors.js';
+
+/**
+ * Searches and returns the lowest common ancestor of two given nodes.
+ *
+ * @param {Node} nodeA First node.
+ * @param {Node} nodeB Second node.
+ * @returns {Node|null} Lowest common ancestor of both nodes or `null` if nodes do not have a common ancestor.
```
Reinmar: The same here. It can be an `HTMLElement` or `DocumentFragment`.

src/dom/getcommonancestor.js
```diff
@@ -0,0 +1,32 @@
...
+	if ( nodeA == nodeB ) {
+		return nodeA;
+	}
+
+	const ancestorsA = getAncestors( nodeA );
+	const ancestorsB = getAncestors( nodeB );
+
+	const minLength = Math.min( ancestorsA.length, ancestorsB.length );
+
+	for ( let i = minLength - 1; i >= 0; i-- ) {
```
Reinmar: Why not iterating from the roots down and checking when the branches?

scofalik: Premature optimization :D.

Because I thought that this way will be more efficient, however this may be dicdated by how I used it. The first usage was to find LCA of two "touching" Text nodes, so I supposed that it will have more common ancestors from top than uncommon ancestors from below.

Yes, now you can bash me.

src/dom/getancestors.js
scofalik: It can be `Node` (it returns itself), it can be `DocumentFragment` or `Document`. I'll update.

src/dom/getancestors.js
scofalik: In fact it's `Node`.

src/dom/getcommonancestor.js
scofalik: In fact it's `{Node|DocumentFragment|Document|null}`

src/dom/getcommonancestor.js
Reinmar: Yes, it's definitely a premature optimization which lead to a code which is harder to understand. So please implement this in the clearer way, especially that both arrays are created beforehand so 99.9% of the job is already done ;).

src/dom/getcommonancestor.js
Reinmar: Why `Node`? And how `Document`?

src/dom/getancestors.js
```diff
@@ -0,0 +1,21 @@
...
+export default function getAncestors( node ) {
+	const nodes = [];
+
+	while ( node ) {
+		nodes.unshift( node );
+		node = node.parentNode;
+	}
+
+	return nodes;
+}
```
pjasiun: You disappointed me, my dear Padawan :(

```js
export default function* getAncestors( node ) {
	while ( node ) {
		yield node;
		node = node.parentNode;
	}
}
```

scofalik: It cannot be done this way because I need to start from root.

Reinmar: I'd like to have this consistent, but I must agree... in many, many scenarios you'll want to start searching from the root. And this array should not be long anyway, so an array will be more handy.

src/dom/getcommonancestor.js
Reinmar: WAT... that's surprising – `document.documentElement.parentNode` returns `document`. I wouldn't expect that at all, so I guess that `getAncestors` should only return elements and doc fragments.

scofalik:
Done. I hope you will be satisfied now :)

src/dom/getcommonancestor.js
scofalik: `Node` - because it returns itself if `nodeA == nodeB`. And `nodeA` can be a `Text` node.

src/dom/getancestors.js
```diff
@@ -3,16 +3,22 @@
...
- * @returns {Array.<Node>} Array of given `node` parents.
+ * @param {Node} node DOM node.
+ * @returns {Array.<Node|DocumentFragment>} Array of given `node` parents.
  */
 export default function getAncestors( node ) {
 	const nodes = [];
 
-	while ( node ) {
+	// Do not return Document object since it's not a `Node` and we are interested in `Node`s (and `DocumentFragment`s).
+	while ( node && !( node instanceof Document ) ) {
```
Reinmar: This is not going to work with other frames. Please use `node.nodeType`.

scofalik:
I used `nodeType` and added a test.

PR #98, opened 2016-11-07
t/97: uid() helper yields non–unique values when imported a couple of times

### LINKED ISSUE #97: The uid helper yields non–unique ids since switching to Webpack/Karma
Webpack builds the `uid.js` module a couple of times and because of that, the **generated UIDs are non–unique in the scale of the whole project anymore**, i.e. they collide if the util has been imported in 2 different modules. 

The problem revealed itself in `DOMEmitterMixin/EmitterMixin` pair because both modules import and use `uid()` helper and collision between the values led to strange behaviors when specific initial conditions were met before running tests.

The safest/fastest/easies solution is importing v4's `getUniqueId()` algorithm https://github.com/ckeditor/ckeditor-dev/blob/master/core/tools.js#L514.

oleq:
Fixes #97.

### DIFF
--- a/src/uid.js
+++ b/src/uid.js
@@ -11,10 +11,12 @@
  * @memberOf utils
  * @returns {Number} A number representing the id.
  */
-export default ( () => {
-	let next = 1;
+export default () => {
+	let uuid = 'e'; // Make sure that id does not start with number.
 
-	return () => {
-		return next++;
-	};
-} )();
+	for ( let i = 0; i < 8; i++ ) {
+		uuid += Math.floor( ( 1 + Math.random() ) * 0x10000 ).toString( 16 ).substring( 1 );
+	}
+
+	return uuid;
+};
--- a/src/uid.js
+++ b/src/uid.js
@@ -9,7 +9,7 @@
  *
  * @function
  * @memberOf utils
- * @returns {Number} A number representing the id.
+ * @returns {String} A string representing the id.
  */
 export default () => {
 	let uuid = 'e'; // Make sure that id does not start with number.
--- a/tests/uid.js
+++ b/tests/uid.js
@@ -12,9 +12,17 @@ describe( 'utils', () => {
 			let id2 = uid();
 			let id3 = uid();
 
-			expect( id1 ).to.be.a( 'number' );
-			expect( id2 ).to.be.a( 'number' ).to.not.equal( id1 ).to.not.equal( id3 );
-			expect( id3 ).to.be.a( 'number' ).to.not.equal( id1 ).to.not.equal( id2 );
+			expect( id1 ).to.be.a( 'string' );
+			expect( id2 ).to.be.a( 'string' ).to.not.equal( id1 ).to.not.equal( id3 );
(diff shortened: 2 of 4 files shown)


Reinmar:
```
  LabeledInput
    constructor()
      child components
        label
          ✖ should have na unique for attribute
            Chrome 54.0.2840 (Mac OS X 10.12.1)
          AssertionError: expected 'ck-input-ecca556263a3ed53511eb1aa524c981a8' to match /^ck-input-[0-9]+$/
              at Context.it (tests/ui-default/labeledinput/labeledinput.js:111:47)

        input
          ✖ should have an unique id attribute
            Chrome 54.0.2840 (Mac OS X 10.12.1)
          AssertionError: expected 'ck-input-e49bc30ba3834724da02b9db1effdc13c' to match /^ck-input-[0-9]+$/
              at Context.it (tests/ui-default/labeledinput/labeledinput.js:125:46)
```

It's temporary, right?

Reinmar:
Besides, you also changed the return type. This will need to be reflected in ckeditor5-ui.

src/uid.js
```diff
@@ -11,10 +11,12 @@
  * @memberOf utils
  * @returns {Number} A number representing the id.
```
Reinmar: Wrong return type.

tests/uid.js
```diff
@@ -12,9 +12,17 @@ describe( 'utils', () => {
...
 			let id3 = uid();
 
-			expect( id1 ).to.be.a( 'number' );
-			expect( id2 ).to.be.a( 'number' ).to.not.equal( id1 ).to.not.equal( id3 );
-			expect( id3 ).to.be.a( 'number' ).to.not.equal( id1 ).to.not.equal( id2 );
+			expect( id1 ).to.be.a( 'string' );
+			expect( id2 ).to.be.a( 'string' ).to.not.equal( id1 ).to.not.equal( id3 );
+			expect( id3 ).to.be.a( 'string' ).to.not.equal( id1 ).to.not.equal( id2 );
+
+			expect( id1[ 0 ] ).to.match( /[a-z]/ );
```
Reinmar: Could we match from `^` to `$`? Otherwise this test makes little sense.

oleq: Fixed.

src/uid.js
oleq: Fixed.

oleq:
As for https://github.com/ckeditor/ckeditor5-utils/pull/98#issuecomment-258816191, let's merge it along with https://github.com/ckeditor/ckeditor5-ui/issues/95 to avoid further conflicts.

As for https://github.com/ckeditor/ckeditor5-utils/pull/98#issuecomment-258816695, there were no docs to be updated because there are none concerning props/attrs generated using `uid()` helper.

Reinmar:
> As for #98 (comment), there were no docs to be updated because there are none concerning props/attrs generated using uid() helper.

At least this: https://github.com/ckeditor/ckeditor5-ui/blob/ebeef8933e5ae3bcb1a7d8dd0924112967f8d354/src/domemittermixin.js#L278

PR #102, opened 2016-11-15
t/88: Allowed event rename on delegation and delegation of all events

### LINKED ISSUE #88: It should be possible to rename event on delegation.
It should be possible to rename event on delegation. That's it.

### LINKED ISSUE #101: It should be possible to delegate all emitter events
In some cases, like in the UI when there's a deep structure which communicates using events (i.e. list item -> list -> panel -> dropdown) and the topmost component (list item) gains a new event that should propagate down to the dropdown, it's quite annoying that each event must be manually delegated as in `listItem.delegate( 'new-event' ).to( list )` then `list.delegate( 'new-event' ).to( panel )`, etc.

Delegation of all events would simplify the code.

oleq:
Closes #88.
Closes #101.

### DIFF
--- a/src/emittermixin.js
+++ b/src/emittermixin.js
@@ -275,14 +275,14 @@ const EmitterMixin = {
 		// Delegate event to other emitters if needed.
 		if ( this._delegations ) {
 			const destinations = this._delegations.get( event );
-			const passThruDestinations = this._delegations.get( '*' );
+			const passAllDestinations = this._delegations.get( '*' );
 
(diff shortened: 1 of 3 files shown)


oleq:
I decided `foo.delegate( '*' ).to( bar )` is better than `foo.delegate().to( bar )` because there will be no confusion when it comes to `foo.stopDelegating( '*' )` vs. `foo.stopDelegating()`, which should have different meanings.

PR #106, opened 2016-12-02
t/105: Moved rect utilities from BalloonPanelView

### LINKED ISSUE #105: Move rect utilities from BalloonPanelView
They should land in `/dom`.

A follow-up of https://github.com/ckeditor/ckeditor5-ui-default/issues/98.

oleq:
Closes #105.

### DIFF
--- a/src/dom/position.js
+++ b/src/dom/position.js
@@ -183,7 +183,7 @@ function getAbsoluteRectCoordinates( { left, top } ) {
 /**
  * Target with respect to which the `element` is to be positioned.
  *
- * @member {HTMLElement|Range} module:utils/dom/position~Options#target
+ * @member {HTMLElement|Range|ClientRect} module:utils/dom/position~Options#target
  */
 
 /**
@@ -197,7 +197,7 @@ function getAbsoluteRectCoordinates( { left, top } ) {
  * When set, the algorithm will chose position which fits the most in the
  * limiter's bounding rect.
  *
- * @member {HTMLElement|Range} module:utils/dom/position~Options#limiter
+ * @member {HTMLElement|Range|ClientRect} module:utils/dom/position~Options#limiter
  */
 
 /**
--- a/src/dom/position.js
+++ b/src/dom/position.js
@@ -34,12 +34,15 @@ export function getOptimalPosition( { element, target, positions, limiter, fitIn
 
 	// If there are no limits, just grab the very first position and be done with that drama.
 	if ( !limiter && !fitInViewport ) {
-		[ bestPositionName, { left, top } ] = getPositionRect( positions[ 0 ], targetRect, elementRect );
+		[ bestPositionName, { left, top } ] = getPosition( positions[ 0 ], targetRect, elementRect );
 	} else {
 		const limiterRect = limiter && new Rect( limiter );
 		const viewportRect = fitInViewport && Rect.getViewportRect();
+		const bestPosition = getBestPosition( positions, targetRect, elementRect, limiterRect, viewportRect );
 
-		[ bestPositionName, { left, top } ] = getBestPositionRect( positions, targetRect, elementRect, limiterRect, viewportRect );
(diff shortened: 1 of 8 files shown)


Reinmar:
`Branches     : 99.52% ( 413/415 )`

R-.

Reinmar:
Firefox doesn't have `ClientRect` global so the whole `rect.js` fails. Therefore, better use duck typing for `isClientRect` and `isElement` (in this case considering future iframe use).

Reinmar:
Actually, we should have `isElement` and `isRange` in `utils/dom/`. I guess that the sufficient way to check whether something is an element is to check that it's an object and its `nodeType == Node.ELEMENT`. `isRange` should also avoid doing `instanceof` check, so perhaps `Object.prototype.toString.apply( range ) == '[object Range]'` would be sufficient.

src/dom/position.js
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * The `getOptimalPosition` helper options.
+ *
+ * @interface module:utils/dom/position~Options
+ */
+
+/**
+ * Element that is to be positioned.
+ *
+ * @member {HTMLElement} module:utils/dom/position~Options#element
```
Reinmar: I don't think you need to specify the whole path. Just the `#element` should be fine (but please do check).

src/dom/position.js
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * An object describing the position in `position: absolute` coordinate
+ * system.
+ *
+ * @typedef {Object} module:utils/dom/position~Position
+ *
+ * @property {Number} top Top position offset.
+ * @property {Number} left Left position offset.
+ * @property {String} name Name of the position.
```
Reinmar: And that's what?

src/dom/rect.js
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * A helper class representing a `ClientRect` object, e.g. value returned by
+ * the native `object.getBoundingClientRect()` method. Provides a set of methods
+ * to manipulate the rect and compare it against other `Rect` instances.
+ */
+export default class Rect {
+	/**
+	 * Creates an instance of rect.
+	 *
+	 * 		// Rect of an HTMLElement.
```
Reinmar: Something broke here with indentation.

src/dom/position.js
oleq: Positions are named. This is useful because as the `getOptimalPosition` util selects the best position, some changes to the DOM may be needed, i.e. the position needs a proper class to be displayed (like an arrow in `BalloonPanelView` which changes with each position).

So, in fact, the knowledge about position geometry (`top`, `left`) is not enough. We need to know, which of the positioning functions passed to https://github.com/ckeditor/ckeditor5-ui-default/blob/t/131/src/balloonpanel/balloonpanelview.js#L145-L150 has been chosen.

That's why positioning functions are named, like https://github.com/ckeditor/ckeditor5-ui-default/blob/t/131/src/balloonpanel/balloonpanelview.js#L207-L211.

----

TBH, I don't like it either but I couldn't find any other way to simplify this API. I mean, to pass a number of functions, get the output data out of one of them (the best one) and to know precisely which function returned this output data. Ideas?

Reinmar: I know that, I didn't mean that it's wrong. I think we can live with it (TBH, I don't have energy to try to find a better solution because this one isn't that bad :D). I just wanted this to be better documented. A single example in this module would do enough.

Reinmar: Actually, the only place where position names look odd is https://github.com/ckeditor/ckeditor5-ui-default/blob/t/131/src/balloonpanel/balloonpanelview.js#L207. But it's reasonable there too. You can both, access the returned value satisfy the interface.

Reinmar: Actually 2, you could mention there that these functions returns objects implementing this specific interface, just to make it clear.

Reinmar:
Great job. I really like the new utils!

PR #111, opened 2016-12-19
t/108a: 	Implemented global module to create stubbable access to global DOM objects

### LINKED ISSUE #108: Failing getOptimalPosition tests on master
https://travis-ci.org/ckeditor/ckeditor5/builds/181701427

A follow-up of https://github.com/ckeditor/ckeditor5-ui-default/issues/137.

oleq:
Closes #108.

### DIFF
--- a/src/dom/global.js
+++ b/src/dom/global.js
@@ -16,11 +16,11 @@
  *
  *		import global from 'ckeditor5/utils/dom/global.js';
  *
- *		const window = global.window;
- *
  *		// This stub will work for any code using global module.
  *		testUtils.sinon.stub( global, 'window', {
  *			innerWidth: 10000
  *		} );
+ *
+ *		console.log( global.window.innerWidth );
  */
 export default { window, document };
--- a/src/dom/global.js
+++ b/src/dom/global.js
@@ -12,13 +12,15 @@
 /**
  * A helper (module) giving an access to the global DOM objects such as `window` and
  * `document`. Accessing these objects using this helper allows easy and bulletproof
- * testing, i.e. stubbing native properties
+ * testing, i.e. stubbing native properties:
  *
  *		import global from 'ckeditor5/utils/dom/global.js';
  *
  *		const window = global.window;
  *
- * 		// This stub will work for any code using global module in tests.
- *		testUtils.sinon.stub( window, 'innerWidth', 10000 );
+ *		// This stub will work for any code using global module.
+ *		testUtils.sinon.stub( global, 'window', {
+ *			innerWidth: 10000
+ *		} );
  */
-export default { document, window };
+export default { window, document };
--- a/src/dom/global.js
+++ b/src/dom/global.js
@@ -1,0 +1,24 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* globals window, document */
+
+/**
+ * @module utils/dom/global
+ */
+
+/**
+ * A helper (module) giving an access to the global DOM objects such as `window` and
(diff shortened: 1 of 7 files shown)


src/dom/global.js
```diff
@@ -0,0 +1,24 @@
...
+ * testing, i.e. stubbing native properties
+ *
+ *		import global from 'ckeditor5/utils/dom/global.js';
+ *
+ *		const window = global.window;
+ *
+ * 		// This stub will work for any code using global module in tests.
+ *		testUtils.sinon.stub( window, 'innerWidth', 10000 );
+ */
+export default { document, window };
```
oleq: I wonder if that could be done better. ATM it's

```js
import global from './global.js';
const window = global.window;
testUtils.sinon.stub( window, 'innerWidth', 10000 );
```

Maybe some way to have `import { window } from './global.js';`?

tests/dom/global.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* global window, document */
+
+import { window as globWindow, document as globDocument } from 'ckeditor5/utils/dom/global.js';
```
oleq: I hope bundlers will handle it properly. Otherwise, it's a simple matter of `const nativeDomWindow = window` before importing and then `expect( window ).to.equal( nativeDomWindow );`

src/dom/global.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/**
+ * A helper (module) giving an access to the global `document` DOM object.
+ * Accessing this object using this helper allows easy and bulletproof
+ * testing, i.e. stubbing native properties:
+ *
+ *		import { document } from 'ckeditor5/utils/dom/global.js';
+ *
+ * 		// This stub will work for any code using global module in tests.
```
Reinmar: Indentation.

src/dom/global.js
```diff
@@ -0,0 +1,34 @@
...
+export { document };
+
+/**
+ * A helper (module) giving an access to the global `window` DOM object.
+ * Accessing this object using this helper allows easy and bulletproof
+ * testing, i.e. stubbing native properties:
+ *
+ *		import { window } from 'ckeditor5/utils/dom/global.js';
+ *
+ * 		// This stub will work for any code using global module in tests.
```
Reinmar: Indentation.

tests/dom/global.js
Reinmar: I don't think that it may cause any problems.

src/dom/global.js
```diff
@@ -0,0 +1,34 @@
...
+ * A helper (module) giving an access to the global `document` DOM object.
+ * Accessing this object using this helper allows easy and bulletproof
+ * testing, i.e. stubbing native properties:
+ *
+ *		import { document } from 'ckeditor5/utils/dom/global.js';
+ *
+ * 		// This stub will work for any code using global module in tests.
+ *		testUtils.sinon.stub( document, 'dir', 'rtl' );
+ */
+export { document };
```
Reinmar: WAIT... what are you actually doing here? Exporting document as document... That doesn't make sense and I don't think it may actually work. Since you're exporting exactly the same object, then stubbing properties on it won't work (in case of window props).

tests/dom/global.js
Reinmar: I take back my previous comment. I didn't read this PR carefully before writing it. Now I have no idea why all this works at all.

oleq:
>  there's some quirk which make them work.

Probably, yes. I might have bulletproofed tests even more by replacing `window.scrollX =...` with `stinon.stub( window, 'scrollX', ... )` and that's why it works.

I guess my changeset changed nothing but this thing :D

PR #118, opened 2017-01-10
Changed: Make EmitterMixin#_listeningTo and _emitterId private

### LINKED ISSUE #114: EmitterMixin#_listeningTo should not be stringified.
Anything that mixes `EmitterMixin` and listens to something, gets `_listeningTo` property. This is also true, of course, for `LiveRange`. Such `LiveRange`, when stringified to JSON, has that property. This is incorrect for `LiveRange` and will probably be incorrect for anything that uses `EmitterMixin`.

This needs to be fixed, and here are two approaches:

1. Specify `toJSON` for each class mixing `EmitterMixin`. In that custom `toJSON` remove `_listeningTo`. This is bad because we need to remember about it each time.
2. Make `_listeningTo` not enumerable. I think this is a better approach, especially since this is hidden, private property.

scofalik:
Fixes #114.

### DIFF
--- a/src/emittermixin.js
+++ b/src/emittermixin.js
@@ -159,11 +159,11 @@ const EmitterMixin = {
 
 		emitters = this[ _listeningTo ];
 
-		if ( !emitter._emitterId ) {
-			emitter._emitterId = uid();
+		if ( !_getEmitterId( emitter ) ) {
+			_setEmitterId( emitter );
 		}
 
-		emitterId = emitter._emitterId;
+		emitterId = _getEmitterId( emitter );
 
 		if ( !( emitterInfo = emitters[ emitterId ] ) ) {
 			emitterInfo = emitters[ emitterId ] = {
@@ -199,7 +199,7 @@ const EmitterMixin = {
 	 */
 	stopListening( emitter, event, callback ) {
 		let emitters = this[ _listeningTo ];
-		let emitterId = emitter && emitter._emitterId;
+		let emitterId = emitter && _getEmitterId( emitter );
 		let emitterInfo = emitters && emitterId && emitters[ emitterId ];
 		let eventCallbacks = emitterInfo && event && emitterInfo.callbacks[ event ];
 
@@ -364,32 +364,6 @@ const EmitterMixin = {
 				destinations.delete( emitter );
 			}
 		}
-	},
-
-	/**
-	 * Emitter's unique id.
-	 *
-	 * **Note:** `_emitterId` can be set only once.
-	 *
-	 * @protected
-	 * @type {String}
-	 */
-	get _emitterId() {
-		return this[ _emitterId ];
-	},
-
-	/**
-	 * Emitter's unique id.
-	 *
-	 * **Note:** `_emitterId` can be set only once.
-	 *
-	 * @protected
-	 * @type {String}
-	 */
-	set _emitterId( id ) {
-		if ( !this[ _emitterId ] ) {
-			this[ _emitterId ] = id;
-		}
 	}
 };
 
@@ -410,6 +384,31 @@ export function _getEmitterListenedTo( listeningEmitter, listenedToEmitterId ) {
 	}
 
 	return null;
+}
+
+/**
+ * Sets emitter's unique id.
+ *
+ * **Note:** `_emitterId` can be set only once.
+ *
+ * @protected
+ * @param {module:utils/emittermixin~EmitterMixin} emitter Emitter for which id will be set.
+ * @param {String} [id] Unique id to set. If not passed, random unique id will be set.
+ */
+export function _setEmitterId( emitter, id ) {
+	if ( !emitter[ _emitterId ] ) {
+		emitter[ _emitterId ] = id || uid();
+	}
+}
+
+/**
+ * Returns emitter's unique id.
+ *
+ * @protected
+ * @param {module:utils/emittermixin~EmitterMixin} emitter Emitter which id will be returned.
+ */
+export function _getEmitterId( emitter ) {
+	return emitter[ _emitterId ];
 }
 
 // Gets the internal `_events` property of the given object.
--- a/src/emittermixin.js
+++ b/src/emittermixin.js
@@ -11,6 +11,12 @@ import EventInfo from './eventinfo';
 import uid from './uid';
 import priorities from './priorities';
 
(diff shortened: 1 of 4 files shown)


scofalik:
It would be nice to followup this with an issue in which we would (try to) get rid off `ck-expando` thing.

Reinmar:
Not sure where, but CC is missing: https://travis-ci.org/ckeditor/ckeditor5-utils/builds/190660925#L1413

Reinmar:
It was 10% previously: https://travis-ci.org/ckeditor/ckeditor5-utils/builds/189164026#L1408

src/emittermixin.js
```diff
@@ -11,6 +11,12 @@ import EventInfo from './eventinfo';
 import uid from './uid';
 import priorities from './priorities';
 
+// Symbol for private `_listeningTo` property.
+const _listeningTo = Symbol();
```
Reinmar: Symbols should be named.

src/emittermixin.js
```diff
@@ -357,6 +367,48 @@ const EmitterMixin = {
...
+
+	/**
+	 * Returns an `EmitterMixin` instance with given `id`, which this emitter is listening to. Returns `null` if
+	 * this emitter do not listen to any emitter identified by `id`.
+	 *
+	 * @protected
+	 * @param {String} id Unique emitter id.
+	 * @returns {EmitterMixin|null}
+	 */
+	_getEmitterListenedTo( id ) {
```
Reinmar: Please do not add more methods to the emitter mixin *object*. Add them to the module. Simply export them from this module and replace `this` with `subject` parameter.

The more methods we add to the object, the bigger chance that it's going clash or something. This is a mixin which should safely work with any class. Besides, it's also uglier.

src/emittermixin.js
```diff
@@ -357,6 +367,48 @@ const EmitterMixin = {
...
 		}
+	},
+
+	/**
+	 * Returns an `EmitterMixin` instance with given `id`, which this emitter is listening to. Returns `null` if
+	 * this emitter do not listen to any emitter identified by `id`.
+	 *
+	 * @protected
+	 * @param {String} id Unique emitter id.
+	 * @returns {EmitterMixin|null}
```
Reinmar: Wrong return type.

src/emittermixin.js
```diff
@@ -357,6 +367,48 @@ const EmitterMixin = {
...
+			return this[ _listeningTo ][ id ].emitter;
+		}
+
+		return null;
+	},
+
+	/**
+	 * Emitter's unique id.
+	 *
+	 * **Note:** `_emitterId` can be set only once.
```
Reinmar: So maybe create only a getter function which sets it if it's not set?

src/emittermixin.js
Reinmar: You can also remove these comments then.

src/emittermixin.js
scofalik: This will not work if you want to set an exact `_emitterId`. The question is whether we really need this functionallity.

src/emittermixin.js
```diff
@@ -150,14 +156,18 @@ const EmitterMixin = {
...
+		if ( !this[ _listeningTo ] ) {
+			this[ _listeningTo ] = {};
 		}
 
-		if ( !( emitterId = emitter._emitterId ) ) {
-			emitterId = emitter._emitterId = uid();
+		emitters = this[ _listeningTo ];
+
+		if ( !emitter._emitterId ) {
+			emitter._emitterId = uid();
```
Reinmar: Yyy.... you sure that it should be `_emitterId`? We were supposed to make it a symbol.

scofalik: This got complicated in the process.

`_emitterId` is available through setter and getter (which are protected). If `_emitterId` is only a private property I should not write `emitter[ _emitterId ]` because then, from "this" emitter I am reaching `_emitterId` of `emitter` (different object).

That's why I introduced getter and setter which are protected. But to "hide" the property itself (so it's not enumerable) I still kept `_emitterId` `Symbol`.

scofalik: Another reason to introduce getter/setter is that we need it in `ProxyEmitter`, of course. Earlier I had `_setEmitterId` and `_getEmitterId` methods. Using them would be perhaps more clear, but it's silly to have such methods when those should be getters.

scofalik: So, to sum up, `Symbol` is used to save the property in the object, then "non-symbol" getter and setter are used to access the property.

Reinmar: Ah... right. So if you fix the other thing (setting the symbol by a separate function, not a class getter), then it will look ok.

src/emittermixin.js
Reinmar: But is anyone using this? I mean - is there a code which needs to set an id?

scofalik: `ProxyEmitterMixin`

Reinmar: Hmm... thinking...

Reinmar: KK, here's explanation why we need the setter now: https://github.com/ckeditor/ckeditor5-utils/issues/120#issuecomment-271881520. Let's simplify this in the followup.

src/emittermixin.js
scofalik: I don't understand what you mean in last comment, but I guess that for now it's fine ?

scofalik:
I fixed few things.

Reinmar:
UI tests fail after these changes.

```
✖ 2 tests failed

FAILED TESTS:
  createListDropdown
    constructor()
      ✖ listens to view#isOpen and reacts to DOM events (valid target)
        Chrome 55.0.2883 (Mac OS X 10.12.2)
      TypeError: Cannot convert undefined or null to object
          at Function.keys (<anonymous>)
          at Context.it (node_modules/ckeditor5-ui/tests/dropdown/list/createlistdropdown.js:9648:19)

      ✖ listens to view#isOpen and reacts to DOM events (invalid target)
        Chrome 55.0.2883 (Mac OS X 10.12.2)
      TypeError: Cannot convert undefined or null to object
          at Function.keys (<anonymous>)
          at Context.it (node_modules/ckeditor5-ui/tests/dropdown/list/createlistdropdown.js:9672:19)
```

Sooory.

PR #124, opened 2017-01-27
T/116

ma2ciek:
Resolves https://github.com/ckeditor/ckeditor5-utils/issues/116.

Translation service can't use 'define' function because of the https://github.com/ckeditor/ckeditor5-dev/issues/59 issue.

### DIFF
--- a/src/locale.js
+++ b/src/locale.js
@@ -55,14 +55,14 @@ export default class Locale {
 	 * @private
 	 */
 	_t( str, values ) {
-		let trasnaltedString = translate( this.lang, str );
+		let translatedString = translate( this.lang, str );
 
 		if ( values ) {
-			trasnaltedString = trasnaltedString.replace( /\%(\d+)/g, ( match, index ) => {
+			translatedString = translatedString.replace( /\%(\d+)/g, ( match, index ) => {
 				return ( index < values.length ) ? values[ index ] : match;
 			} );
 		}
 
-		return trasnaltedString;
+		return translatedString;
 	}
 }
--- a/src/locale.js
+++ b/src/locale.js
@@ -7,6 +7,8 @@
  * @module utils/locale
  */
 
+import { translate } from './translation-service';
+
 /**
  * Represents the localization services.
  */
@@ -53,10 +55,14 @@ export default class Locale {
 	 * @private
 	 */
 	_t( str, values ) {
+		let trasnaltedString = translate( this.lang, str );
+
 		if ( values ) {
-			str = str.replace( /\%(\d+)/g, ( match, index ) => ( index < values.length ) ? values[ index ] : match );
+			trasnaltedString = trasnaltedString.replace( /\%(\d+)/g, ( match, index ) => {
+				return ( index < values.length ) ? values[ index ] : match;
+			} );
 		}
 
-		return str;
+		return trasnaltedString;
 	}
 }
--- a/src/translation-service.js
+++ b/src/translation-service.js
@@ -7,66 +7,59 @@
  * @module utils/translation-service
  */
 
-const translations = {};
+let dictionaries = {};
 
 /**
- * Merges package translations to existing ones.
- * These translations can be used later with {@link module:utils/translations-service~translate translate}
+ * Adds package translations to existing ones.
+ * These translations can be used later with {@link module:utils/translations-service~translate translate}.
  *
  *		add( 'pl', {
  *			'OK': 'OK',
  *			'Cancel [context: reject]': 'Anuluj'
- *		} );OK
+ *		} );
  *
- * @param {String} lang
- * @param {Object.<String, Object>} packageDictionary
- * @returns undefined
+ * @param {String} lang Target language.
+ * @param {Object.<String, String>} translations translations which will be added to a dictionary.
  */
-export function add( lang, packageDictionary ) {
-	if ( !( lang in translations ) ) {
-		translations[ lang ] = {};
-	}
-
-	const dictionary = translations[ lang ];
-
-	for ( const translationKey in packageDictionary ) {
-		const translation = packageDictionary[ translationKey ];
-		dictionary[ translationKey ] = translation;
-	}
+export function add( lang, translations ) {
+	dictionaries[ lang ] = dictionaries[ lang ] || {};
+	Object.assign( dictionaries[ lang ], translations );
 }
 
 /**
- * Translates string if the translation of the string was previously added using {@link module:utils/translations-service~add add}.
- * Otherwise returns original (English) sentence.
+ * Translates string if the translation of the string was previously added using {@link module:utils/translations-service~add add}
+ * (multi-language mode). When no translation is defined in the dictionary or the dictionary doesn't exists it returns original string
+ * without the '[context: ]' (development mode and single-language mode). In single-language mode the strings are without the contexts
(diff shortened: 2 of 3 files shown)


src/locale.js
```diff
@@ -53,10 +55,14 @@ export default class Locale {
 	 * @private
 	 */
 	_t( str, values ) {
+		let trasnaltedString = translate( this.lang, str );
```
Reinmar: Typo.

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+/**
+ * @module utils/translation-service
+ */
+
+const translations = {};
+
+/**
+ * Merges package translations to existing ones.
```
Reinmar: It doesn't merge anything – it simply adds translations. Do not overcomplicate docs ;).

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+ * These translations can be used later with {@link module:utils/translations-service~translate translate}
```
Reinmar: Missing period.

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ *		add( 'pl', {
+ *			'OK': 'OK',
+ *			'Cancel [context: reject]': 'Anuluj'
+ *		} );OK
```
Reinmar: Typo.

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @param {String} lang
```
Reinmar: Missing param description

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+ * @param {Object.<String, Object>} packageDictionary
```
Reinmar: You don't know if this is a package's dictionary. These are translations and can come from any source.

This option also misses description and the type is wrong too (values are not objects).

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+ * @returns undefined
```
Reinmar: No need to document that;

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+export function add( lang, packageDictionary ) {
+	if ( !( lang in translations ) ) {
+		translations[ lang ] = {};
+	}
+
+	const dictionary = translations[ lang ];
+
+	for ( const translationKey in packageDictionary ) {
+		const translation = packageDictionary[ translationKey ];
+		dictionary[ translationKey ] = translation;
```
Reinmar: Overcomplicated. Please merge these two lines.

Besides, `Object.assign()` should be better.

tests/translation-service.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import { translate, add, _clear } from '../src/translation-service';
+
+describe( 'translationService', () => {
```
Reinmar: Wrong name – it's `translation-service`.

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/**
+ * Translates string if the translation of the string was previously added using {@link module:utils/translations-service~add add}.
+ * Otherwise returns original (English) sentence.
+ *
+ *		translate( 'pl', 'Cancel [context: reject]' );
+ *
+ * @param {String} lang Translation language.
+ * @param {String} translationKey Sentence which is going to be translated.
```
Reinmar: It doesn't have to be a sentence.

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+/**
+ * Translates string if the translation of the string was previously added using {@link module:utils/translations-service~add add}.
+ * Otherwise returns original (English) sentence.
+ *
+ *		translate( 'pl', 'Cancel [context: reject]' );
+ *
+ * @param {String} lang Translation language.
```
Reinmar: Incorrect description. It's not "translation's language" (it doesn't make any sense). It's more like "Target language." or "Language which dictionary will be used.".

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+export function translate( lang, translationKey ) {
+	if ( !existTranslationKey( lang, translationKey ) ) {
+		return translationKey.replace( / \[context: [^\]]+\]$/, '' ) ;
+	}
+
+	return translations[ lang ][ translationKey ];
+}
+
+function existTranslationKey( lang, translationKey ) {
```
Reinmar: `hasTranslation()`

Reinmar: or `hasTranslationFor()`

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ *		translate( 'pl', 'Cancel [context: reject]' );
+ *
+ * @param {String} lang Translation language.
+ * @param {String} translationKey Sentence which is going to be translated.
+ * @returns {String} Translated sentence.
+ */
+export function translate( lang, translationKey ) {
+	if ( !existTranslationKey( lang, translationKey ) ) {
+		return translationKey.replace( / \[context: [^\]]+\]$/, '' ) ;
```
Reinmar: Typo.

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+ * Otherwise returns original (English) sentence.
+ *
+ *		translate( 'pl', 'Cancel [context: reject]' );
+ *
+ * @param {String} lang Translation language.
+ * @param {String} translationKey Sentence which is going to be translated.
+ * @returns {String} Translated sentence.
+ */
+export function translate( lang, translationKey ) {
+	if ( !existTranslationKey( lang, translationKey ) ) {
```
Reinmar: I'm not sure whether this code makes any sense (and if it does, it requires comments). Please explain me how's it going to work in the dev env (where no translations are defined and `t()` is called with the original strings) and in the situation when translation keys were replaced with chosen language. AFAICT, it's going to work in both of these situations, but I'd like to have confirmation and comments for that. Perhaps the best way for them will be the documentation of this function which now is incorrect ("returns original (English) sentence" is not entirely true in the second mentioned situation).

Please also remember to explain the not yet implemented support for multiple languages.

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
...
+	);
+}
+
+/**
+ * Clears translations for test purpose.
+ *
+ * @protected
+ */
+export function _clear() {
+	for ( const key in translations ) {
```
Reinmar: `translations = {};` would be easier.

src/translation-service.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * @license Copyright (c) 2003-2016, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module utils/translation-service
+ */
+
+const translations = {};
```
Reinmar: Maybe `dictionaries`? You later call values of this object `dictionary` (L30).

src/translation-service.js
ma2ciek: Ow, I missed that after the concept has changed.

PR #127, opened 2017-03-02
Added default configuration parameter for Config

### LINKED ISSUE #126: Config – an easy way to load the default config
As posted here - https://github.com/ckeditor/ckeditor5-core/pull/72#issuecomment-283611499 - we need to introduce default parameters for `Config`.

It is required for support `Editor.build` property (https://github.com/ckeditor/ckeditor5-core/issues/67).

pomek:

Feature: Added default configuration parameter for `Config`. Closes #126.

---

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -34,7 +34,7 @@ export default class Config {
 
 		// Set default configuration.
 		if ( defaultConfigurations ) {
-			this._setObjectToTarget( this._config, defaultConfigurations, true );
+			this.define( defaultConfigurations );
 		}
 	}
 
--- a/src/config.js
+++ b/src/config.js
@@ -18,7 +18,7 @@ export default class Config {
 	 *
 	 * @param {Object} [configurations] The initial configurations to be set.
 	 */
-	constructor( configurations ) {
+	constructor( configurations, defaultConfigurations ) {
 		/**
 		 * Store for the whole configuration.
 		 *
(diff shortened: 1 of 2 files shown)


src/config.js
```diff
@@ -31,6 +31,11 @@ export default class Config {
 		if ( configurations ) {
 			this._setObjectToTarget( this._config, configurations );
 		}
+
+		// Set default configuration.
+		if ( defaultConfigurations ) {
+			this._setObjectToTarget( this._config, defaultConfigurations, true );
```
Reinmar: Why not `define()`? It's a role of this method to set the default values.

pomek: Fixed in `9bd0d4b`.

Reinmar:
Please add a test for a deeper config objects.

pomek:
Done in 4783f38.

tests/config.js
```diff
@@ -41,9 +41,75 @@ describe( 'Config', () => {
...
+			expect( config.get( 'bar' ) ).to.equal( 2 );
+		} );
+
+		it( 'should work with deeper objects', () => {
+			const defaultConfig = {
+				a: {
+					first: 1,
+					second: 2
+				},
+				'b.foo.first': 1,
```
Reinmar: Not like this, like:

```js
b: {
   foo: {
       first: 1
   }
}
```

This is supported just by coincident :D

pomek: Fixed in 2e7eab7. :D

Reinmar:
You forgot about API docs but I'll add them. R+ for the rest.

PR #131, opened 2017-03-14
T/125: Imported bindTo method from ViewCollection

### LINKED ISSUE #125: Move ViewCollection.bind() to utils.Collection
Since it's a very useful feature to bind two collections, especially if you can do custom mapping, it should be moved to the `utils.Collection` to let users bind any collections, not only `ViewCollection`. 

Note that you can bind any observable variable, but not any observable collection, what is inconsistent.

The problem is that the binding checks if the parameter is the instance of `View` to check if it should be used or if the parameter is a creator function (https://github.com/ckeditor/ckeditor5-ui/blob/6b7fd4adfe6ee99fbdf84bfc4940f10ed78493b8/src/viewcollection.js#L227-L246). To solve it we could split the `as` function to separate function, what would make the code simpler and less magic:

```js
collection1.bindTo( collection2 ).byWrappingInto( MyViewClass );
collection1.bindTo( collection2 ).using( ( object ) => object.propery );
```

Note that in the future we might introduce another popular mapping helper, like for the last line from the example above:

```
collection1.bindTo( collection2 ).withProperty( 'property' );
```

oleq:

Other: Imported bindTo method ViewCollection. Closes #125.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -55,6 +55,15 @@ export default class Collection {
 		 * @member {String}
 		 */
 		this._idProperty = options && options.idProperty || 'id';
+
+		/**
+		 * A helper mapping external items from bound collection ({@link #bindTo})
+		 * and actual items of the collection.
+		 *
+		 * @protected
+		 * @member {Map}
+		 */
+		this._boundItemsMap = new Map();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -271,6 +280,140 @@ export default class Collection {
...
+			 *
+			 * @static
+			 * @param {Function|String} CallbackPropertyOrClass When specified as a callback, the callback
+			 * produces the items. When the class constructor is passed, the factory of that class is initialized.
+			 * When the string is provided, the property value is used to create the binding.
+			 */
+			using: ( CallbackPropertyOrClass ) => {
+				let createItem;
+
+				if ( CallbackPropertyOrClass.toString().match( /^class/ ) ) {
```
oleq: I wonder if there's any better way to check this.

As a second thought, this could be `/^class\s+/` because theoretically, there could be some object with custom `toString()` implementation returning `'classicCreator'` string or something like that.

pjasiun: This is why I proposed to avoid magic and simple have to separate method: 1 for callbacks and 1 for classes.

pjasiun: Even in your proposal there was `as` what is fine for me.

oleq: OK, I'm gonna split it then into as/using pair.

pjasiun:
I think that to make this feature fully useful two additional tickets should be fixed:
 - two-way collections binding (more-less https://github.com/ckeditor/ckeditor5-ui/issues/110),
 - skipping elements (https://github.com/ckeditor/ckeditor5-ui/issues/92).

tests/collection.js
```diff
@@ -512,6 +512,177 @@ describe( 'Collection', () => {
...
+					items.remove( 1 );
+					expect( collection ).to.have.length( 1 );
+					expect( collection.get( 0 ).value ).to.equal( 'foo' );
+
+					items.remove( 0 );
+					expect( collection ).to.have.length( 0 );
+				} );
+			} );
+		} );
+	} );
```
pjasiun: I miss a test for 
```
if ( item.label == 'foo' ) {
	return new FooClass( item );
} else {
	return new BarClass( item );
}
```

PR #135, opened 2017-03-15
T/130: Added 'first' function.

### LINKED ISSUE #130: `first` util to get the first item from the generator
It's basically `generator.next().value;`, but:

```js
first( doc.selection.getSelectedBlocks() );
```
seems to be more readlable them:

```js
doc.selection.getSelectedBlocks().next().value;
```

I know we have `nth`, but it's a common case when you wants to get the first value and we could have a dedicated method for it.

ma2ciek:

Feature: Added 'first' function. Closes #130.

### DIFF
--- a/src/first.js
+++ b/src/first.js
@@ -8,7 +8,7 @@
  */
 
 /**
- * Returns first item of an `iterable`.
+ * Returns first item of the given `iterable`.
  *
  * @param {Iterable.<*>} iterable
  * @returns {*}
@@ -16,9 +16,9 @@
 export default function first( iterable ) {
 	const iteratorItem = iterable.next();
 
-	if ( !iteratorItem.done ) {
-		return iteratorItem.value;
+	if ( iteratorItem.done ) {
+		return null;
 	}
 
-	return null;
+	return iteratorItem.value;
 }
--- a/src/first.js
+++ b/src/first.js
@@ -1,0 +1,24 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module utils/first
+ */
+
+/**
+ * Returns first item of an `iterable`.
+ *
+ * @param {Iterable.<*>} iterable
+ * @returns {*}
+ */
+export default function first( iterable ) {
+	const iteratorItem = iterable.next();
+
+	if ( !iteratorItem.done ) {
+		return iteratorItem.value;
+	}
+
+	return null;
+}
--- a/tests/first.js
+++ b/tests/first.js
@@ -1,0 +1,22 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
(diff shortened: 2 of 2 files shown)


src/first.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Returns first item of an `iterable`.
+ *
+ * @param {Iterable.<*>} iterable
+ * @returns {*}
+ */
+export default function first( iterable ) {
+	const iteratorItem = iterable.next();
+
+	if ( !iteratorItem.done ) {
```
pjasiun: The positive condition with the "early return" will be more natural.

pjasiun: I mean, the code without `!` is usually a little bit simpler to read than with it.

Second, usual construction is:
```
    if (!somethingIsWrong)
        return;

    // Do something default
```

src/first.js
```diff
@@ -0,0 +1,24 @@
...
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module utils/first
+ */
+
+/**
+ * Returns first item of an `iterable`.
```
Reinmar: "of *the* iterable" or even "of the given iterable"

tests/first.js
```diff
@@ -0,0 +1,22 @@
...
+
+			expect( first( iterator ) ).to.equal( 11 );
+		} );
+
+		it( 'should return null if iterator is empty', () => {
+			const iterator = [][ Symbol.iterator ]();
+
+			expect( first( iterator ) ).to.be.null;
+		} );
+	} );
```
Reinmar: I miss a test what happens if you use `first()` twice for one iterable. Since we talked about this it's usually good to also confirm it with a test. Plus, it may clarify it for others what's the behaviour.

tests/first.js
```diff
@@ -0,0 +1,22 @@
...
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import first from '../src/first';
+
+describe( 'utils', () => {
+	describe( 'first', () => {
+		it( 'should return first item', () => {
+			const iterator = [ 11, 22 ][ Symbol.iterator ]();
```
pjasiun: Don't do too much in one line. The code above might be not easily readable for someone. I prefer:

```js
const collection = [ 11, 22 ];
const iterator = collection[ Symbol.iterator ]();
```

Reinmar: Good point. It took me a couple of secs to understand what the `[][ Symbol.iterator()` does :P.

src/first.js
ma2ciek: I'll update similar comment in the `nth()` function description.

tests/first.js
ma2ciek: Done.

src/first.js
ma2ciek: For only one condition more logical is "if something than do something". Here there're two return statements so probably your version is a little bit more elegant.

Reinmar:
This would be a better message:

> Feature: Introduced the `first()` function (gets the first item from an iterable). Closes #130.

PR #138, opened 2017-03-15
Added tests and docs to the edge cases in the 'nth' function.

### LINKED ISSUE #137: Improve 'nth()' function tests and documentation
As mentioned #136 and #130 there're some edge cases which are not tested and documented. Especially differences between taking iterators and generators as the second argument

ma2ciek:

Internal: Added tests and docs to the edge cases in the 'nth' function. Closes #137.

### DIFF
--- a/src/nth.js
+++ b/src/nth.js
@@ -8,14 +8,16 @@
  */
 
 /**
- * Returns `nth` (starts from `0` of course) item of an `iterable`.
+ * Returns `nth` (starts from `0` of course) item of the given `iterable`.
+ * Consumes all items of the generator.
+ * Consumes all items up to the index of the iterator, including the returned item.
  *
  * @param {Number} index
  * @param {Iterable.<*>} iterable
  * @returns {*}
  */
 export default function nth( index, iterable ) {
-	for ( let item of iterable ) {
+	for ( const item of iterable ) {
 		if ( index === 0 ) {
 			return item;
 		}
--- a/src/nth.js
+++ b/src/nth.js
@@ -10,7 +10,9 @@
 /**
  * Returns `nth` (starts from `0` of course) item of the given `iterable`.
  * Consumes all items of the generator.
- * Consumes all items up to the index of the iterator, including the returned item.
+ * Consumes all items up to the index of the iterator that doesn't implement the `return()` method, (e.g. Array Iterator).
+ * It's necessary to implement the `return()` method to close the iterator after the return statement in the `nth` function is invoked.
+ * @see http://www.ecma-international.org/ecma-262/7.0/#sec-iteratorclose.
  *
  * @param {Number} index
  * @param {Iterable.<*>} iterable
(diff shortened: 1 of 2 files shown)


tests/nth.js
```diff
@@ -8,26 +8,43 @@ import nth from '../src/nth';
...
+		it( 'should stop inside the given iterator', () => {
+			const collection = [ 11, 22, 33 ];
+			const iterator = collection[ Symbol.iterator ]();
+
+			nth( 0, iterator );
+
+			expect( iterator.next().value ).to.equal( 22 );
+		} );
+
+		function *getGenerator() {
```
Reinmar: It is a generator which returns an iterator. So it should stay being `getIterator()`.

src/nth.js
```diff
@@ -8,14 +8,16 @@
  */
 
 /**
- * Returns `nth` (starts from `0` of course) item of an `iterable`.
+ * Returns `nth` (starts from `0` of course) item of the given `iterable`.
+ * Consumes all items of the generator.
+ * Consumes all items up to the index of the iterator, including the returned item.
```
Reinmar: This looks like a duplicated sentence. So how does it work finally?

ma2ciek: it consumes the generator and it doesn't consume the iterator if the iterator doesn't implement the `return() method`, like e.g. Array Iterator. http://www.ecma-international.org/ecma-262/7.0/#sec-iteratorclose. I can add it to the comments to clarify what's going on.

tests/nth.js
ma2ciek: Generator implements the iterator protocol, but the iterator and generator objects aren't the same and behave differently in some situations.

src/nth.js
ma2ciek: These tests are showing the difference: https://github.com/ckeditor/ckeditor5-utils/pull/138/files#diff-7b7a935b05c5a95c53b4e26cadf83040R30

tests/nth.js
Reinmar: OK, I didn't know about that. So if you want to differentiate that, then the a better name would be `generateValues()` or `createGenerator()` **AND** it such things must be documented somewhere. Note that none of us would understand the real meaning of this change.

Reinmar: Hm... Or maybe I'm overexaggerating or even wrong. I checked that generator is an object, the function is called "generator function" (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator). So you're right – it can be `getGenerator()` because it actually returns a generator. My bad.

Reinmar: Previously I thought that that function is a generator and it returns an iterator.

ma2ciek: For me both `getGenerator` and `createGenerator` seem good. `Generator` is a little bit unintuitive term.

ma2ciek: > Previously I thought that that function is a generator and it returns an iterator.

Reinmar: Let's keep `getGenerator()`. After I understood the proper naming of these objects it all makes sense to me.

ma2ciek:
Docs for the `nth` are done.

PR #140, opened 2017-03-20
T/132: Two–way data binding between Collection instances

### LINKED ISSUE #132: Two–way binding between Collection instances
A follow-up of https://github.com/ckeditor/ckeditor5-ui/issues/110, since `bindTo` method has been moved from `ViewCollection` to utils `Collection` in https://github.com/ckeditor/ckeditor5-utils/issues/125.

oleq:

Feature: Two–way data binding between Collection instances. Closes #132.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -61,18 +61,18 @@ export default class Collection {
 		 * and actual items of this collection.
 		 *
 		 * @protected
-		 * @member {Map}
+		 * @member {WeakMap}
 		 */
-		this._bindToExternalToInternalMap = new Map();
+		this._bindToExternalToInternalMap = new WeakMap();
 
 		/**
 		 * A helper mapping items of this collection to external items of a bound collection
 		 * ({@link #bindTo}).
 		 *
 		 * @protected
-		 * @member {Map}
+		 * @member {WeakMap}
 		 */
-		this._bindToInternalToExternalMap = new Map();
+		this._bindToInternalToExternalMap = new WeakMap();
 
 		/**
 		 * A collection instance this collection is bound to as a result
--- a/src/collection.js
+++ b/src/collection.js
@@ -66,6 +66,15 @@ export default class Collection {
 		this._bindToExternalToInternalMap = new Map();
 
 		/**
+		 * A helper mapping items of this collection to external items of a bound collection
+		 * ({@link #bindTo}).
+		 *
+		 * @protected
+		 * @member {Map}
+		 */
+		this._bindToInternalToExternalMap = new Map();
+
+		/**
 		 * A collection instance this collection is bound to as a result
 		 * of calling {@link #bindTo} method.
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -57,13 +57,30 @@ export default class Collection {
...
+		this._bindToExternalToInternalMap = new Map();
+
+		/**
+		 * A helper mapping items of this collection to external items of a bound collection
+		 * ({@link #bindTo}).
+		 *
+		 * @protected
+		 * @member {Map}
+		 */
+		this._bindToInternalToExternalMap = new Map();
```
pjasiun: Can't it be a `WeaMap`? Does not look like a real collection more like mapping and `WeaMap` wound be safer for mapping because it will be cleared even if one forget to call `clear`. I see you iterate over this map only in tests what seems to be too deeply testing, checking internal, what is a bad smell. Can't test check only the content of collections without checking internal?

pjasiun: Also, it's not clear what these variable are used for. What is `external items`? What is `internal`? A longer description would be useful since there variable seems to be essential for the binding.

tests/collection.js
```diff
@@ -728,6 +783,235 @@ describe( 'Collection', () => {
...
+
+				assertItems( collectionA, [ 4, 6, 8 ], [ [ 2, 4 ], [ 3, 6 ], [ 4, 8 ] ] );
+				assertItems( collectionB, [ 2, 3, 4 ], [ [ 4, 2 ], [ 6, 3 ], [ 8, 4 ] ] );
+				assertItems( collectionC, [ -2, -3, -4 ], [ [ 2, -2 ], [ 3, -3 ], [ 4, -4 ] ] );
+
+				collectionC.add( { v: -1000 } );
+
+				assertItems( collectionA, [ 4, 6, 8 ], [ [ 2, 4 ], [ 3, 6 ], [ 4, 8 ] ] );
+				assertItems( collectionB, [ 2, 3, 4 ], [ [ 4, 2 ], [ 6, 3 ], [ 8, 4 ] ] );
+				assertItems( collectionC, [ -2, -3, -4, -1000 ], [ [ 2, -2 ], [ 3, -3 ], [ 4, -4 ] ] );
```
pjasiun: Aaa! Magic numbers. @scofalik would love it ;) But beeing serious, I wanted to write that these assertions should be split into separate `assertItems` and `assertMapping` what would make them more readable, but, as I wrote, I don't think mapping should be tested, but only the result of the mapping.

src/collection.js
oleq: I'll check this out. What you suggest is probably a good idea.

tests/collection.js
oleq: Maybe you're right. This helped me a lot during development but shouldn't necessary remain in the tests.

PR #145, opened 2017-04-07
Emitter#stopListening() and Emitter#off() improvements

### LINKED ISSUE #144: Should EmitterMixin.stopListening() off all listeners (those added through on() too)?
While working on https://github.com/ckeditor/ckeditor5-engine/pull/904 @scofalik noticed that `stopListening()` is not removing listeners added using `on()`.

I assume it was such a case, right @scofalik?

```js
a.on( 'x',  callback );

a.stopListening();
```

If this is true, then it means that you can't easily clean all listeners that you registered. My first thought was that it makes this method useless.

However, as I found that I already explained in https://github.com/ckeditor/ckeditor5-design/issues/142 (sic! :D), `this.on()` is to be used by a class internally. So if class X needs to listen to its own events, it can use `this.on()`. This means that together with the class its listeners will be gone because no more events will be fired as well.

However, again, this is a theory, because we don't enforce the rule that you can fire events only on yourself. So someone may mistakenly still fire events on an already destroyed instance of class X and the callbacks will be executed. This increases a chance of a bug.

Taken that, I'd say that:

* either `stopListening()` should `off()` all listeners added by `on()`,
* or we should always use `listenTo()` and remove `on()` (alternatively, `on()` should use `listenTo()` internally).

I wouldn't totally remove `on()` because its shortness is useful. Having to always do `a.listenTo( a, ...)` would be cumbersome. Also, we use `on()` mostly in tests and half of those cases where we use it source are incorrect (`listenTo()` should be used with a proper subject).

So, I'm either for making `on()` a shortcut to `a.listenTo( a )` or `stopListening()` offing all events. In the former case we need to figure out what `off()` should do. In the latter, nothing changes AFAICT.

WDYT?

scofalik:

Fix: `utils.Emitter#off()` called without arguments will stop executing all callbacks added by `utils.Emitter#on()`. Additionally, `utils.Emitter#stopListening()` called without arguments will also call `utils.Emitter#off()` thus stop all callbacks added by `utils.Emitter#on()`. Closes #144.

### DIFF
--- a/src/emittermixin.js
+++ b/src/emittermixin.js
@@ -99,12 +99,12 @@ const EmitterMixin = {
 	},
 
 	/**
-	 * Stops executing the callback on the given event. It can be used at different levels:
+	 * Removes the callback for the given event. It can be used at different levels:
 	 *
-	 * To stop executing given callback, with given context.
-	 * To stop executing given callback.
-	 * To stop executing all callbacks for given event.
-	 * To stop executing all callbacks for this emitter.
+	 * * To stop executing given callback, with given context.
+	 * * To stop executing given callback.
+	 * * To stop executing all callbacks for given event.
+	 * * To stop executing all callbacks for this emitter.
 	 *
 	 * @method #off
 	 * @param {String} [event] The name of the event.
@@ -116,13 +116,13 @@ const EmitterMixin = {
 		if ( !event ) {
 			const events = getEvents( this );
 
-			for ( let event in events ) {
+			for ( const event in events ) {
 				this.off( event );
 			}
 		} else {
 			const lists = getCallbacksListsForNamespace( this, event );
 
-			for ( let callbacks of lists ) {
+			for ( const callbacks of lists ) {
 				for ( let i = 0; i < callbacks.length; i++ ) {
 					if ( !callback || callbacks[ i ].callback == callback ) {
 						if ( !context || context == callbacks[ i ].context ) {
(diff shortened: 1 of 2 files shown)


Reinmar:
I pushed a couple of improvements, but unfortunately, I find this module too messy. We're just asking for troubles. You can even see that the currently implemented solution is a hack by how it stands out from the rest of the code: https://github.com/ckeditor/ckeditor5-utils/pull/145/files#diff-1b61a350b30716326bca3d6e478665d8R219

For me, `this.on()` must use `this.listenTo( this, ... )`. This may need some refactoring because today the `on()` method is the base one, while `listenTo()` uses it. For me, it should be the opposite.

I know that we could live with the proposed solution for months, but for me this is too important piece of the code to have any doubts about how it works. We can see that it already leaked and that all of us were surprised what happened (even though I found this in the past). So, since the entire editor bases on this mechanism I'm for keep it as clean as possible.

Reinmar:
R-, but I'm open for discussion ofc :P

scofalik:
I, too, hate this whole mixin. But this PR blocks this: https://github.com/ckeditor/ckeditor5-engine/pull/904 . Maybe we could refactor it in separate issue?

Reinmar:
> But this PR blocks this: ckeditor/ckeditor5-engine#904 . Maybe we could refactor it in separate issue?

This issue isn't critical so we're not in hurry. Having it open and having this one open will force us to get back to this topic sooner.

scofalik:
I've made some improvements on branch `144-b` but they are not ready yet. I will be tinkering with this in "meantimes".

Reinmar:
#190 replaces this PR so I'm closing this one.

PR #147, opened 2017-04-07
t/146: DomEmitterMixin#stopListening ignores useCapture.

### LINKED ISSUE #146: DomEmitterMixin#stopListening ignores useCapture
In result, event listeners are not stopped at all, if created with capturing.

From https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener:
> Specifies whether the EventListener to be removed is registered as a capturing listener or not. If this parameter is absent, a default value of false is assumed.
If a listener is registered twice, one with capture and one without, remove each one separately. Removal of a capturing listener does not affect a non-capturing version of the same listener, and vice versa.

oleq:

Fixed: DomEmitterMixin#stopListening should respect listeners attached with useCapture. Closes #146.

### DIFF
--- a/src/dom/emittermixin.js
+++ b/src/dom/emittermixin.js
@@ -196,7 +196,7 @@ extend( ProxyEmitter.prototype, EmitterMixin, {
 			return;
 		}
 
-		const domListener = this._createDomListener( event );
+		const domListener = this._createDomListener( event, !!options.useCapture );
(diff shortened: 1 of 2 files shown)


Reinmar:
Please, write changelog messages as "should". It clearly indicates what happens after the change.

PR #149, opened 2017-04-18
T/148: getOptimalPosition utility should consider limiter ancestors with CSS overflow. Closes #148.

### LINKED ISSUE #148: getOptimalPosition utility should consider limiter ancestors with overflow
> Additionally getOptimalPosition() could check all the ancestors of the limiter which have overflow different than visible and intersect all their rects one by one up to window to find the real visible area which is available to position the BalloonPanel. That would, in most cases, make the limiter configuration obsolete.

A follow-up of https://github.com/ckeditor/ckeditor5-ui/issues/173.

oleq:

Fix: The getOptimalPosition utility should consider limiter ancestors with CSS overflow. Closes #148.


src/dom/rect.js
```diff
@@ -180,6 +194,50 @@ export default class Rect {
...
+		const obj = this._obj;
+		let visibleRect = this.clone();
+
+		// There's no ancestor to crop <body> with the overflow.
+		if ( obj != global.document.body ) {
+			let parent = obj.parentNode || obj.commonAncestorContainer;
+
+			// Check the ancestors all the way up to the <body>.
+			while ( parent && parent != global.document.body ) {
+				const parentOverflow = global.window.getComputedStyle( parent ).overflow;
```
oleq: I wonder if checking each parent as below (`const intersectionRect = visibleRect.getIntersection( parentRect );`) wouldn't be actually faster than using this `window.getComputedStyle`. Will benchmark.

oleq: ...and I was right. `window.getComputedStyle` is slow. Twice the performance after https://github.com/ckeditor/ckeditor5-utils/pull/149/commits/24e1d149b0953fa7e63ff92a43e0f6f0ac08da65.

src/dom/rect.js
```diff
@@ -40,6 +40,20 @@ export default class Rect {
...
 	 */
 	constructor( obj ) {
+		/**
+		 * The object this rect is for.
+		 *
+		 * @protected
+		 * @readonly
+		 * @member {HTMLElement|Range|ClientRect|module:utils/dom/rect~Rect|Object} #_obj
+		 */
+		Object.defineProperty( this, '_obj', {
```
oskarwrobel: Why `defineProperty` ?
Maybe `_context` instead of `_obj`?

tests/dom/position.js
```diff
@@ -159,6 +159,31 @@ describe( 'getOptimalPosition()', () => {
 				name: 'left'
 			} );
 		} );
+
+		// https://github.com/ckeditor/ckeditor5-utils/issues/148
+		it( 'should return coordinates (#3)', () => {
+			const overflowedAncestor = getElement( {
```
oskarwrobel: Webstorm says `Local variable 'overflowedAncestor' is redundant`.

tests/dom/position.js
```diff
@@ -159,6 +159,31 @@ describe( 'getOptimalPosition()', () => {
...
+				top: 100,
+				left: 0,
+				bottom: 110,
+				right: 10,
+				width: 10,
+				height: 10
+			}, {
+				overflow: 'scroll'
```
oskarwrobel: Looks like `overflow: scroll` is no longer needed by this test but OTOH it better explains this test case.

oskarwrobel:
I left few comments but works 👌

src/dom/rect.js
oleq: `defineProperty` to set `enumerable` `false`. Without it iterating over the `Rect` would also return `_obj` and make some pieces of code more complex.

As for `_obj`, maybe not `_context` because it makes people think about `this` and stuff but `_from`?

tests/dom/position.js
oleq: This is odd. It's used.

oskarwrobel: This warning is about
```
const overflowedAncestor = getelement();

limiter.parentNode = overflowedAncestor;
```
vs
```
limiter.parentNode = getelement();
```
but it's a detail.

PR #154, opened 2017-05-04
t/153: Rect utility should work for collapsed DOM Ranges. Closes #153.

### LINKED ISSUE #153: Rect utility should work for collapsed DOM Ranges
ATM, collapsed Range at the beginning of the block has bounding client rect full of zeros. `getClientRects()[ 0 ]` solves the issue but [when the block is empty, even this method returns an empty array](https://github.com/ckeditor/ckeditor5-ui-default/pull/91#issuecomment-251344158). A smarter approach is needed, probably `startContainer.getBoundingClientRect` with a small correction of width.

A follow-up of https://github.com/ckeditor/ckeditor5-link/issues/46#issuecomment-298966477.

oleq:

Fix: Rect utility should work for collapsed DOM Ranges. Closes #153.

### DIFF
--- a/src/dom/rect.js
+++ b/src/dom/rect.js
@@ -54,11 +54,39 @@ export default class Rect {
 			enumerable: false
 		} );
 
-		if ( isElement( source ) || isRange( source ) ) {
-			source = source.getBoundingClientRect();
-		}
+		// Acquires all the rect properties from the passed source.
+		//
+		// @private
+		// @param {ClientRect|module:utils/dom/rect~Rect|Object} source}
+		const setProperties = source => {
+			rectProperties.forEach( p => this[ p ] = source[ p ] );
+		};
 
-		rectProperties.forEach( p => this[ p ] = source[ p ] );
+		if ( isElement( source ) ) {
+			setProperties( source.getBoundingClientRect() );
+		} else if ( isRange( source ) ) {
+			// Use getClientRects() when the range is collapsed
+			// https://github.com/ckeditor/ckeditor5-utils/issues/153
+			if ( source.collapsed ) {
+				const rects = source.getClientRects();
+
+				if ( rects.length ) {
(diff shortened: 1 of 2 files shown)


src/dom/rect.js
```diff
@@ -54,11 +54,39 @@ export default class Rect {
...
 		} );
 
-		if ( isElement( source ) || isRange( source ) ) {
-			source = source.getBoundingClientRect();
-		}
+		// Acquires all the rect properties from the passed source.
+		//
+		// @private
+		// @param {ClientRect|module:utils/dom/rect~Rect|Object} source}
+		const setProperties = source => {
```
Reinmar: Use function definitions.

oleq: What about `this` then? I think it would only complicate the code.

src/dom/rect.js
```diff
@@ -54,11 +54,39 @@ export default class Rect {
...
+			rectProperties.forEach( p => this[ p ] = source[ p ] );
+		};
 
-		rectProperties.forEach( p => this[ p ] = source[ p ] );
+		if ( isElement( source ) ) {
+			setProperties( source.getBoundingClientRect() );
```
Reinmar: I don't like how many times `setProperties` is executed in this file. Couldn't the rect be assigned to some value and later on the values be set once?

oleq: It used to be that way. But since the new case arrived, where`width` must be set to `0` and the output of `getBoundingClientRect`/`getClientRects` is read–only, I would need to use an intermediate object to write client rect's properties to it (in a loop, `Object.assign` fails) and then alter the `witdth`, which doesn't look pretty. So I decided this looks better.

Reinmar: KK. Makes sense. I didn't now it's read only.

PR #158, opened 2017-05-16
t/157: getOptimalPosition should accept a target defined as a function

### LINKED ISSUE #157: Rect constructor should accept a function as a source
A follow-up of https://github.com/ckeditor/ckeditor5-ui/issues/227 and https://github.com/ckeditor/ckeditor5-ui/issues/195.

oleq:

Other: `getOptimalPosition` utility should accept the target option defined as a function. Closes #157.

### DIFF
--- a/src/dom/position.js
+++ b/src/dom/position.js
@@ -76,6 +76,12 @@ import getPositionedAncestor from './getpositionedancestor';
  * @returns {module:utils/dom/position~Position}
  */
 export function getOptimalPosition( { element, target, positions, limiter, fitInViewport } ) {
+	// If the {@link module:utils/dom/position~Options#target} is a function, use what it returns.
+	// https://github.com/ckeditor/ckeditor5-utils/issues/157
+	if ( typeof target == 'function' ) {
+		target = target();
+	}
+
 	const positionedElementAncestor = getPositionedAncestor( element.parentElement );
 	const elementRect = new Rect( element );
 	const targetRect = new Rect( target );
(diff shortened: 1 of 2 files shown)


Reinmar:
I don't understand this change. Why can't you pass the result which the function would return straight to `Rect()`'s constructor?

oleq:
It could be [here](https://github.com/ckeditor/ckeditor5-utils/blob/master/src/dom/position.js#L81) as well but I think making it low–level will spare us similar conditions in other places in the future.

Reinmar:
I really don't understand why do need this callback. It does nothing except returning a value. Either I miss something or it's totally unnecessary.

oleq:
OK, to be clear, this change is caused by https://github.com/ckeditor/ckeditor5-ui/pull/233. The pinning algorithm in the `BalloonPanelView`, when configured for a Rect (not a Range or HTMLElement), it goes crazy upon window scroll – the Rect is immutable, and once set as a target, it will prevent the balloon from re–positioning.

So to make things easier, I allowed target configuration as a function, which for each scroll event is called at some level (either Rect's constructor, or `getOptimalPosition`) and returns the Rect which is up to date, allowing correct balloon positioning.

Reinmar:
I'd understand your comment that we need to make the `Rect` live. I don't think this is the right way around, but I also don't see this in your PR. The callback is executed immediately by the constructor so there's no difference between passing a callback to `Rect()` and a real param. 

I'm still confused and now I'm quite sure that this should be avoided.

oleq:
After F2F talk, I moved the fix to `getOptimalPosition`.

PR #163, opened 2017-06-08
Introduced ObservableMixin#decorate() and support for fire()'s return values

### LINKED ISSUE #162: Implement ObservableMixin.decorate()
A common pattern which we use is that in order to make some component's action overridable we implement it using an event. The general flow is like this:

1. The method does only one thing – fires an event (called like that method).
2. The default action of that method is implemented in form of a listener with a default priority.

One can then easily listen to that event, modify arguments, cancel it, perform additional actions, etc.

We always did this manually which means adding a bit of ugly code and is not DRY. When talking about https://github.com/ckeditor/ckeditor5-core/issues/88 we thought that `execute()` could be implemented in the same way but we can't have even more code duplication and the commands' code should be as clean as possible.

Hence, the idea to implement `OM.decorate()` (big simplification of the decorators idea). It will take care of replacing the original method with a method which fires and event and plugs a listener.

## EventInfo#return

Part two – implement `EventInfo#return` so you can return from `fire()`. This will be needed for https://github.com/ckeditor/ckeditor5-engine/issues/963 and it makes no sense to work on these two things separately.

Reinmar:

Feature: Introduced `ObservableMixin#decorate()` and support for setting `EmitterMixin#fire()`'s return value by listeners. Closes #162.

### DIFF
--- a/src/emittermixin.js
+++ b/src/emittermixin.js
@@ -22,14 +22,23 @@ const _emitterId = Symbol( 'emitterId' );
  */
 const EmitterMixin = {
 	/**
-	 * Registers a callback function to be executed when an event is fired. Events can be grouped in namespaces using `:`.
+	 * Registers a callback function to be executed when an event is fired.
+	 *
+	 * Events can be grouped in namespaces using `:`.
 	 * When namespaced event is fired, it additionaly fires all callbacks for that namespace.
 	 *
 	 *		myEmitter.on( 'myGroup', genericCallback );
 	 *		myEmitter.on( 'myGroup:myEvent', specificCallback );
-	 *		myEmitter.fire( 'myGroup' ); // genericCallback is fired.
-	 *		myEmitter.fire( 'myGroup:myEvent' ); // both genericCallback and specificCallback are fired.
-	 *		myEmitter.fire( 'myGroup:foo' ); // genericCallback is fired even though there are no callbacks for "foo".
+	 *
+	 *		// genericCallback is fired.
+	 *		myEmitter.fire( 'myGroup' );
+	 *		// both genericCallback and specificCallback are fired.
+	 *		myEmitter.fire( 'myGroup:myEvent' );
+	 *		// genericCallback is fired even though there are no callbacks for "foo".
+	 *		myEmitter.fire( 'myGroup:foo' );
+	 *
+	 * An event callback can {@link module:utils/eventinfo~EventInfo#stop stop the event} and
+	 * set the {@link module:utils/eventinfo~EventInfo#return return value} of the {@link #fire} method.
 	 *
 	 * @method #on
 	 * @param {String} event The name of the event.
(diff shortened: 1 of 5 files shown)


src/observablemixin.js
```diff
@@ -247,6 +247,76 @@ const ObservableMixin = {
...
+	 *			args[ 0 ] = 3;
+	 *
+	 *			console.log( args[ 1 ] ); // -> 2
+	 *		}, { priority: 'high' } );
+	 *
+	 *		foo.method( 1, 2 ); // -> '3, 2'
+	 *
+	 * @param {String} ...methodNames Names of the methods to decorate.
+	 */
+	decorate( ...methodNames ) {
```
pjasiun: I would keep it simple: we don't expect many methods to be decorated at the same time, so I don't find it needed to add handling multiple methods at the same time by this function. Also, it would be a problem if we will want to add some parameters to it in the future.

src/emittermixin.js
```diff
@@ -135,6 +135,9 @@ const EmitterMixin = {
 	 * the priority value the sooner the callback will be fired. Events having the same priority are called in the
 	 * order they were added.
 	 * @param {Object} [options.context] The object that represents `this` in the callback. Defaults to the object firing the event.
+	 * @returns By default the method returns `undefined`. However, the return value can be changed by listeners
+	 * through modification of the {@link module:utils/emitterinfo~EmitterInfo#return}'s value (the event info
+	 * is the first param of every callback).
```
pjasiun: It should be docs for `fire` not for `listenTo`. However, a comment here about how callback can set the return value would be fine.

src/observablemixin.js
Reinmar: I actually took that into consideration. I think that if we'd add options it would become an object anyway:

```js
decorate( { name: 'foo', how: 'bar' } )
```

But perhaps you're right that support for decorating multiple methods at once is an overkill. I'll remove that.

src/emittermixin.js
Reinmar: Haha. Good catch.

Reinmar:
When working on tests for ckeditor5-core I stumbled upon a case where I tried to spy on `command.execute` and the `command` was created using `new Command()`. This got me an ugly error because `command.execute` did exist, but the `originalMethod` property inside `observablemixin.js` didn't. That's because I forgot to implement the `execute()` method.

To avoid such issues I'll make `decorate()` throw an error.

Reinmar:
Back on review.

Reinmar:
One last doubt – the `decorate()` method could be moved to `EmitterMixin` because it only requires the event system. It semantically makes more sense in *Observable*Mixin but it means that classes may need to start mixing it instead of `EmitterMixin` to get this functionality (it was a case for `DataController`).

szymonkups:
Wrong button :D

PR #172, opened 2017-07-07
t/168: Provided a static `Rect#getDomRangeRects` method for external usage

### LINKED ISSUE #168: Provide a static Rect#getDomRangeRects method for external usage
ATM there's no way to obtain all DOM Range rects e.g. when the range spans multiple lines of text using the `Rect` class. While it's possible to iterate over the list returned by `domRange.getClientRects()` and create a new `Rect` instance from each rect, such approach does not cover very specific cases which are **already handled** by the `Rect` helper class e.g. when there's no rect (Safari) etc..

So to use the smart logic of the `Rect` class and avoid code duplication, a static method is required. The very first use case is in https://github.com/ckeditor/ckeditor5-ui/issues/263, where depending on the direction of the seleciton, a proper DOM Range rect must be returned (either the first or the last one). Without the smart `Rect` logic, `ContextualToolbar` fails to show when the selection is collapsed in Safari and in other edge cases.

oleq:

Other: Provided a static `Rect#getDomRangeRects` method for external usage. Closes #168.

### DIFF
--- a/src/dom/rect.js
+++ b/src/dom/rect.js
@@ -55,25 +55,7 @@ export default class Rect {
 		if ( isElement( source ) ) {
 			copyRectProperties( this, source.getBoundingClientRect() );
 		} else if ( isRange( source ) ) {
-			// Use getClientRects() when the range is collapsed.
-			// https://github.com/ckeditor/ckeditor5-utils/issues/153
-			if ( source.collapsed ) {
-				const rects = source.getClientRects();
-
-				if ( rects.length ) {
-					copyRectProperties( this, rects[ 0 ] );
-				}
-				// If there's no client rects for the Range, use parent container's bounding
-				// rect instead and adjust rect's width to simulate the actual geometry of such
-				// range.
-				// https://github.com/ckeditor/ckeditor5-utils/issues/153
-				else {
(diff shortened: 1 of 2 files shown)


Reinmar:
I don't understand these changes :D Waiting for more related PRs to see the reason for it.

src/dom/rect.js
```diff
@@ -55,25 +55,7 @@ export default class Rect {
 		if ( isElement( source ) ) {
 			copyRectProperties( this, source.getBoundingClientRect() );
 		} else if ( isRange( source ) ) {
-			// Use getClientRects() when the range is collapsed.
-			// https://github.com/ckeditor/ckeditor5-utils/issues/153
-			if ( source.collapsed ) {
```
Reinmar: Where did that go?

oleq: It has been unified in `getDomRangeRects`. There's no need to differentiate it anymore.

Reinmar:
> I don't understand these changes :D Waiting for more related PRs to see the reason for it.

@oleq explained this to me that there are some features which want to arbitrarily choose rects from the native range and based on that create `Rect` instances. But if they take one native rect and pass it to `Rect()` then that special logic of `Rect()` constructor which checks whether the range is collapsed etc wasn't executed.

PR #174, opened 2017-08-01
t/173: Created various DOM utilities to allow the view document scroll to the selection

### LINKED ISSUE #173: Create DOM utils to help implementing the viewport scrolling helper
The viewport/ancestors scrolling helper will require certain enhancements to the `Rect` class and possibly other helpers.

A follow–up of https://github.com/ckeditor/ckeditor5-engine/issues/660.

oleq:

Feature: Created various DOM utilities to allow the view document scroll to the selection. Closes #173.

BREAKING CHANGE: The static `Rect.getViewportRect` method has been removed. Use `new Rect( window )` instead.

### DIFF
--- a/src/dom/getborderwidths.js
+++ b/src/dom/getborderwidths.js
@@ -10,24 +10,19 @@
 import global from './global';
 
 /**
- * Returns an object containing CSS border withs of a specified `HTMLElement`.
+ * Returns an object containing CSS border widths of a specified HTML element.
  *
  * @param {HTMLElement} element An element which has CSS borders.
  * @param {Object} An object containing `top`, `left`, `right` and `bottom` properties
  * with numerical values of the `border-[top,left,right,bottom]-width` CSS styles.
  */
 export default function getBorderWidths( element ) {
-	const computedStyles = global.window.getComputedStyle( element );
-	const borderWidths = {
-		top: computedStyles.borderTopWidth,
-		right: computedStyles.borderRightWidth,
-		bottom: computedStyles.borderBottomWidth,
-		left: computedStyles.borderLeftWidth
+	const style = global.window.getComputedStyle( element );
+
+	return {
+		top: parseInt( style.borderTopWidth, 10 ),
+		right: parseInt( style.borderRightWidth, 10 ),
+		bottom: parseInt( style.borderBottomWidth, 10 ),
+		left: parseInt( style.borderLeftWidth, 10 )
 	};
-
-	for ( const width in borderWidths ) {
-		borderWidths[ width ] = parseInt( borderWidths[ width ], 10 );
-	}
-
-	return borderWidths;
 }
--- a/src/dom/getborderwidths.js
+++ b/src/dom/getborderwidths.js
@@ -1,0 +1,33 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module utils/dom/getborderwidths
+ */
+
+import global from './global';
+
+/**
+ * Returns an object containing CSS border withs of a specified `HTMLElement`.
+ *
+ * @param {HTMLElement} element An element which has CSS borders.
+ * @param {Object} An object containing `top`, `left`, `right` and `bottom` properties
+ * with numerical values of the `border-[top,left,right,bottom]-width` CSS styles.
+ */
+export default function getBorderWidths( element ) {
+	const computedStyles = global.window.getComputedStyle( element );
+	const borderWidths = {
+		top: computedStyles.borderTopWidth,
+		right: computedStyles.borderRightWidth,
+		bottom: computedStyles.borderBottomWidth,
+		left: computedStyles.borderLeftWidth
+	};
+
+	for ( const width in borderWidths ) {
+		borderWidths[ width ] = parseInt( borderWidths[ width ], 10 );
+	}
+
+	return borderWidths;
+}
--- a/src/dom/rect.js
+++ b/src/dom/rect.js
@@ -250,7 +250,7 @@ export default class Rect {
 	 * Checks if all properties ({@link #top}, {@link #left}, {@link #right},
 	 * {@link #bottom}, {@link #width} and {@link #height}) are the same as in the other `Rect`.
 	 *
-	 * @param {Rect} rect A `Rect` instance to compare with.
+	 * @param {module:utils/dom/rect~Rect} rect A `Rect` instance to compare with.
 	 * @returns {Boolean} `true` when Rects are equal. `false` otherwise.
 	 */
 	isEqual( anotherRect ) {
@@ -340,7 +340,8 @@ export default class Rect {
 	 * Excludes scrollbars and CSS borders from the `Rect`.
 	 *
 	 * * Borders are removed when {@link #_source} is `HTMLElement`.
-	 * * Scrollbars are excluded from `HTMLElements` and {@link #getViewportRect viewport rects}.
+	 * * Scrollbars are excluded from `HTMLElements` and
+	 * {@link module:utils/dom/rect~Rect.getViewportRect viewport rects}.
 	 *
 	 * @private
 	 */
--- a/src/dom/rect.js
+++ b/src/dom/rect.js
@@ -51,7 +51,7 @@ export default class Rect {
 		 *
 		 * @protected
 		 * @readonly
-		 * @member {HTMLElement|Range|ClientRect|module:utils/dom/rect~Rect|object} #_source
+		 * @member {HTMLElement|Range|ClientRect|module:utils/dom/rect~Rect|Object} #_source
 		 */
 		Object.defineProperty( this, '_source', {
 			// If the source is a Rect instance, copy it's #_source.
--- a/src/dom/rect.js
+++ b/src/dom/rect.js
@@ -27,32 +27,31 @@ export default class Rect {
 	 *		// Rect of a DOM Range.
 	 *		const rectB = new Rect( document.getSelection().getRangeAt( 0 ) );
 	 *
+	 *		// Rect of a window (web browser viewport).
+	 *		const rectC = new Rect( window );
+	 *
 	 *		// Rect out of an object.
-	 *		const rectC = new Rect( { top: 0, right: 10, bottom: 10, left: 0, width: 10, height: 10 } );
+	 *		const rectD = new Rect( { top: 0, right: 10, bottom: 10, left: 0, width: 10, height: 10 } );
 	 *
 	 *		// Rect out of another Rect instance.
-	 *		const rectD = new Rect( rectC );
+	 *		const rectE = new Rect( rectD );
 	 *
 	 *		// Rect out of a ClientRect.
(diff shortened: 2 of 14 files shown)


src/dom/getborderwidths.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+/**
+ * @module utils/dom/getborderwidths
+ */
+
+import global from './global';
+
+/**
+ * Returns an object containing CSS border withs of a specified `HTMLElement`.
```
Reinmar: widths*

HTML element

src/dom/getborderwidths.js
```diff
@@ -0,0 +1,33 @@
...
+export default function getBorderWidths( element ) {
+	const computedStyles = global.window.getComputedStyle( element );
+	const borderWidths = {
+		top: computedStyles.borderTopWidth,
+		right: computedStyles.borderRightWidth,
+		bottom: computedStyles.borderBottomWidth,
+		left: computedStyles.borderLeftWidth
+	};
+
+	for ( const width in borderWidths ) {
```
Reinmar: Just use `parseInt()` above and return the object without iterating over it. That will be shorter and faster.

src/dom/rect.js
```diff
@@ -234,21 +247,63 @@ export default class Rect {
 	}
 
 	/**
+	 * Checks if all properties ({@link #top}, {@link #left}, {@link #right},
+	 * {@link #bottom}, {@link #width} and {@link #height}) are the same as in the other `Rect`.
```
Reinmar: are the same -> equal? properties are "the same"  has a bit different meaning than "values of these properties equal ...". It's a detail and shouldn't be confusing, but I found the wording a bit unusual.

src/dom/rect.js
```diff
@@ -234,21 +247,63 @@ export default class Rect {
...
+			if ( this[ prop ] !== anotherRect[ prop ] ) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	/**
+	 * Checks whether a `Rect` fully contains another `Rect` instance.
```
Reinmar: The same issue as everywhere else – I don't think that it makes sense to abuse preformatted text and type names. IMO, this makes reading the descriptions harder because the text is less heterogeneous. Plus, it's semantically incorrect as well because you're not checking `Rect` (type) here but a rect instance.

For me, it should be: "Checks whether this rect fully contains another rect instance."

The same applies to many other places in these docs.

Reinmar: cc @AnnaTomanek

src/dom/rect.js
```diff
@@ -234,21 +247,63 @@ export default class Rect {
...
+	 *
+	 * @param {module:utils/dom/rect~Rect} anotherRect
+	 * @returns {Boolean} `true` if contains, `false` otherwise.
+	 */
+	contains( anotherRect ) {
+		const intersectRect = this.getIntersection( anotherRect );
+
+		if ( !intersectRect || !intersectRect.isEqual( anotherRect ) ) {
```
Reinmar: `return intersectRect && intersectRect.isEqual( anotherRect );`?

src/dom/rect.js
```diff
@@ -234,21 +247,63 @@ export default class Rect {
...
+
+	/**
 	 * Returns a rect of the web browser viewport.
 	 *
 	 * @returns {module:utils/dom/rect~Rect} A viewport rect.
+	 * @param {Boolean} [options.excludeScrollbars] When set `true` the `Rect` will not include
+	 * the scrollbars of the viewport.
 	 */
-	static getViewportRect() {
+	static getViewportRect( options = {} ) {
```
Reinmar: Why does this method look like one big hack? There's something wrong with this method's constructor if you need to override stuff like `_source` and call private methods once an instance was initialised.

src/dom/rect.js
```diff
@@ -234,21 +247,63 @@ export default class Rect {
...
 			bottom: innerHeight,
 			left: 0,
 			width: innerWidth,
 			height: innerHeight
 		} );
+
+		rect._source = global.window;
+
+		if ( options.excludeScrollbars ) {
+			rect._excludeScrollbarsAndBorders();
```
Reinmar: Couldn't this be a public method? It'd make more sense to call it manually, rather than add the `options` param everywhere.

src/dom/scroll.js
```diff
@@ -0,0 +1,164 @@
...
+ * This helper will scroll all `target` ancestors and the web browser viewport to reveal the target to
+ * the user. If the `target` is already visible, nothing will happen.
+ *
+ * @param {HTMLElement|Range} options.target A target, which supposed to become visible to the user.
+ * @param {Number} [options.viewportOffset] An offset from the edge of the viewport (in pixels)
+ * the `target` will be moved by when the viewport is scrolled. It enhances the user experience
+ * by keeping the `target` some distance from the edge of the viewport and thus making it easier to
+ * read or edit by the user.
+ */
+function scrollViewportToShowTarget( { target, viewportOffset = 0 } ) {
```
Reinmar: won't `scrollViewportToShowTarget( { target } )` crash? It will leave `viewportOffset` set to undefined or not? I can never remember;|

src/dom/scroll.js
```diff
@@ -0,0 +1,164 @@
...
+	} );
+
+	// Avoid the situation where the caret is still in the viewport, but totally
+	// at the edge. If it moved beyond the viewport in the next action e.g. due to enter,
+	// the scrolling would move it to the viewportOffset level and it all would look like the
+	// caret visually moved up/down.
+	//
+	// To prevent this, we're checking the rects moved by the viewportOffset to cover
+	// the upper/lower edge of the viewport.
+	const rects = [ targetShiftedUpRect, targetShiftedDownRect ];
```
Reinmar: Yyyy... Are you sure about this method's code? I don't want to analyse it :D

src/dom/scroll.js
```diff
@@ -0,0 +1,164 @@
...
+			excludeScrollbarsAndBorders: true
+		} );
+
+		if ( !parentRect.contains( targetRect ) ) {
+			scrollAncestorToShowTarget( { targetRect, parent, parentRect } );
+		}
+	} while ( ( parent = parent.parentNode ) );
+}
+
+export default {
```
Reinmar: Why exporting this as an object? For testing purposes? Please explain it then. And in fact, it's better to expose these functions separately and then expose a test object with the two of them. So that object will be protected, while the functions will be nice and clean and exposed separately.

oleq: Yes. It's for testing purposes, e.g. to be able to mock `scrollAncestorsToShowTarget()` and check if `scrollViewportToShowTarget()` always calls it first. 

Without the mock, the test becomes ridiculously complex because both viewport and ancestors are to be scrolled and it's fairly obvious that `scrollAncestorsToShowTarget()` is already tested extensively in the other place.

src/dom/scroll.js
```diff
@@ -0,0 +1,164 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* global Node */
+
+/**
+ * @module utils/dom/scroll
```
oleq: There are [lots](https://github.com/ckeditor/ckeditor5-utils/pull/174/files#diff-94487ad070262037a0877767f74a71c1) of them.

src/dom/scroll.js
oleq: It won't. See the [tests](https://github.com/ckeditor/ckeditor5-utils/pull/174/files#diff-94487ad070262037a0877767f74a71c1R236) and [tutorials](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#ES2015_version).

src/dom/scroll.js
Reinmar: OK, they got folded so I missed them.

src/dom/scroll.js
Reinmar: OK, the problems I remembered were with this notation: `function x( a = { x: 1, y: 2 } )`.

src/dom/scroll.js
Reinmar: OK, but you can still code this in a clearer way:

```js
const module = { a, b };

export function a() {
	module.b();
}

export function b() {
	...
}

export module as _test;
```

Thanks to the fact that `a()` uses `b()` through `module` you can still mock the stuff.

Reinmar:
I love the manual test :D The description is unclear (it's too complicated – it's hard to figure out what you're actually supposed to do) but you can play with these elements in various ways.

I'd only do one change – make the squares borders' rounded. That will help in recognizing whether the entire square is visible. Something like 5px will be enough. You can also think about different style which might help. 1px border would also be good.

src/dom/rect.js
oleq: I removed the method and allowed `new Rect( window )`. See the updated docs.

src/dom/rect.js
oleq: I made a public method out of this thing.

src/dom/scroll.js
oleq: Yes. It's a pure magic 😄

src/dom/rect.js
```diff
@@ -26,36 +27,54 @@ export default class Rect {
...
-			writable: false,
+			writable: true,
 			enumerable: false
 		} );
 
 		if ( isElement( source ) ) {
 			copyRectProperties( this, source.getBoundingClientRect() );
 		} else if ( isRange( source ) ) {
 			copyRectProperties( this, Rect.getDomRangeRects( source )[ 0 ] );
+		} else if ( source === global.window ) {
```
Reinmar: What about other windows? Iframes will have their own. This deserves a followup.

PS. The next line is not optimal (you could use `source`).

src/dom/rect.js
```diff
@@ -234,21 +253,67 @@ export default class Rect {
...
+	 * * Scrollbars are excluded from HTML elements and the `window`.
+	 *
+	 * @returns {module:utils/dom/rect~Rect} A rect which has been updated.
+	 */
+	excludeScrollbarsAndBorders() {
+		const source = this._source;
+		let scrollBarWidth, scrollBarHeight;
+
+		if ( source === global.window ) {
+			scrollBarWidth = global.window.innerWidth - global.document.documentElement.clientWidth;
```
Reinmar: Just like I commented above – depending on the main window and document is unsafe.

src/dom/rect.js
```diff
@@ -234,21 +253,67 @@ export default class Rect {
...
+	}
+
+	/**
+	 * Checks whether a rect fully contains another rect instance.
+	 *
+	 * @param {module:utils/dom/rect~Rect} anotherRect
+	 * @returns {Boolean} `true` if contains, `false` otherwise.
+	 */
+	contains( anotherRect ) {
+		const intersectRect = this.getIntersection( anotherRect );
```
Reinmar: What about two rects next to each other?

Reinmar: OK, nvmnd. I didn't ask :D

Reinmar:
And here's the followup: https://github.com/ckeditor/ckeditor5-utils/issues/175.

PR #177, opened 2017-08-15
Remove "anySymbol in smth" statements.

DarthVictor:
Remove "anySymbol in smth" statements. They are incompatible with core-js polyfill.
In fact that's a core-js implementation (Symbol('any string') in AnyObject === true) problem, but it is much more difficult to fix babel-polyfill rather then two lines of code.

Type: Message. Closes #000.

---

### Additional information

### DIFF
--- a/src/observablemixin.js
+++ b/src/observablemixin.js
@@ -197,7 +197,7 @@ const ObservableMixin = {
 	 */
 	unbind( ...unbindAttrs ) {
 		// Nothing to do here if not inited yet.
-		if ( !( attributesSymbol in this ) ) {
+		if ( !this[attributesSymbol] ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


Reinmar:
Hey. Thanks for the PR. 

Would you be so kind and report an issue in this repository explaining the problem? It will be easier to proceed with it. I guess I understand what's the problem but it'd be good to have it in black and white for the future reference.

Besides, we'll need a test(s) which will verify this change. The test could, as far as I understand this case, mock how core-js polyfills symbols and check that the corrected methods work fine.

PS. Why do you load core-js? All browsers supported by CKEditor 5 at this moment don't require the polyfill for ES6. Do you use core-js for something else?

DarthVictor:
> PS. Why do you load core-js? All browsers supported by CKEditor 5 at this moment don't require the polyfill for ES6. Do you use core-js for something else?

Android 4.4 WebView unfortunately needs polyfill. So the problem is simple - CKEditor crashed during initialization. I'll report an issue to the repo in one or two days.

PR #179, opened 2017-08-17
t/178: A warning should show up when a Rect instance is for an HTML element or a DOM range not belonging to a rendered DOM tree

### LINKED ISSUE #178: Rect.getDomRangeRects doesn't support ranges inside the text nodes
Running `documentView.scrollToTheSelection` when selection is inside some text nodes throws the following Error:

`Error: Uncaught TypeError: range.startContainer.getBoundingClientRect is not a function.`

oleq:

Internal: A warning should show up when a Rect instance is for an HTML element or a DOM range not belonging to a rendered DOM tree. Closes #178.

----

Branches in other repositories to silence the warnings:

* https://github.com/ckeditor/ckeditor5-editor-inline/compare/t/ckeditor5-utils/178
* https://github.com/ckeditor/ckeditor5-ui/compare/t/ckeditor5-utils/178
* https://github.com/ckeditor/ckeditor5-engine/compare/t/ckeditor5-utils/178

### DIFF
--- a/src/dom/rect.js
+++ b/src/dom/rect.js
@@ -11,6 +11,7 @@ import global from './global';
 import isRange from './isrange';
 import isElement from '../lib/lodash/isElement';
 import getBorderWidths from './getborderwidths';
+import log from '../log';
 
 /**
  * A helper class representing a `ClientRect` object, e.g. value returned by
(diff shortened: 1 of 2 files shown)


oleq:
I decided against the error and I went with the `log.warn`. An error would mean fixing ~30 tests in various repositories, which does not make sense and is just a waste of time IMO.

Reinmar:
Hm... And did you mute these warnings in all these repositories? That, unfortunately, needs to be done anyway. It's simpler, though.

oleq:
You're right. I'm gonna mute those warnings.

oleq:
Done.

PR #181, opened 2017-08-21
t/ckeditor5/447: KeystrokeHandler should support priorities and proper keystroke canceling

### LINKED ISSUE #180: KeystrokeHandler should support priorities and proper canceling
`KeystrokeHandler` callbacks must have priorities just like `Emitter` events. I will allow us to make sure that no matter the order plugins are loaded, `PluginX`'s callback will always prevail over the `PluginY`'s if their priority is high.

Alo, ATM, if a listener in the KH cancels like
   ```js
   hk.set( 'Tab', ( data, cancel ) => {
   	cancel();
   } );
   ```
   it only causes `preventDefault` and `stopPropagation` in DOM. It does not prevent other `Tab` callbacks from executing and that needs to be fixed if the `PluginX` callback is to be executed but the one belonging to the `PluginY` ignored.

Related issue https://github.com/ckeditor/ckeditor5-link/issues/87.
A follow-up of https://github.com/ckeditor/ckeditor5/issues/447#issuecomment-320698055.

oleq:

Feature: KeystrokeHandler should support priorities and proper keystroke canceling. Closes #180.

### DIFF
--- a/src/keystrokehandler.js
+++ b/src/keystrokehandler.js
@@ -9,6 +9,7 @@
 
 import DomEmitterMixin from './dom/emittermixin';
 import { getCode, parseKeystroke } from './keyboard';
+import priorities from './priorities';
 
 /**
  * Keystroke handler registers keystrokes so the callbacks associated
@@ -19,7 +20,7 @@ import { getCode, parseKeystroke } from './keyboard';
  *
  *		handler.listenTo( emitter );
  *
- *		handler.set( 'ctrl + a', ( keyEventData, cancel ) => {
+ *		handler.set( 'ctrl + a', ( keyEvtData, cancel ) => {
  *			console.log( 'ctrl + a has been pressed' );
  *			cancel();
  *		} );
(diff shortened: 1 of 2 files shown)


src/keystrokehandler.js
```diff
@@ -65,47 +67,48 @@ export default class KeystrokeHandler {
...
+	 * @param {Object} [options={}] Additional options.
+	 * @param {module:utils/priorities~PriorityString|Number} [options.priority='normal'] The priority of the keystroke
+	 * callback. The higher the priority value the sooner the callback will be executed. Keystrokes having the same priority
+	 * are called in the order they were added.
 	 */
-	set( keystroke, callback ) {
+	set( keystroke, callback, options = {} ) {
 		const keyCode = parseKeystroke( keystroke );
-		const callbacks = this._keystrokes.get( keyCode );
+		const priority = priorities.get( options.priority );
```
Reinmar: Is this necessary? Won't `listenTo()` do this automatically?

oleq: Hmmm... it's been a while since I coded this and I don't remember. Try removing it and see if any test fails.

PR #195, opened 2017-10-22
Introduced Collection#first and Collection#last getters

### LINKED ISSUE #194: Getters for first and last item in the Collection
It would be very useful (and easy to do) to have getters for `first`and `last` item in the Collection.

oskarwrobel:

Feature: Introduced `Collection#first` and `Collection#last` getters. Closes #194.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -31,7 +31,7 @@ export default class Collection {
 	 * @param {Object} [options={}] The options object.
 	 * @param {String} [options.idProperty='id'] The name of the property which is considered to identify an item.
 	 */
-	constructor( { idProperty = 'id' } = {} ) {
+	constructor( options = {} ) {
 		/**
 		 * The internal list of items in the collection.
 		 *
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -28,10 +28,10 @@ export default class Collection {
 	/**
 	 * Creates a new Collection instance.
 	 *
-	 * @param {Object} options The options object.
+	 * @param {Object} [options={}] The options object.
 	 * @param {String} [options.idProperty='id'] The name of the property which is considered to identify an item.
 	 */
-	constructor( options ) {
+	constructor( { idProperty = 'id' } = {} ) {
```
ma2ciek: The con of this approach is that you're losing the JSDoc comment to param connection. I'm not sure how JSDoc behaves in such moments.

Reinmar: I'd say that this should be reverted as this isn't part of this ticket anyway ;)

pjasiun: Hi, @Reinmar! How are your holidays?

Reinmar: I actually start them in a couple hours. You can see this by the weather ;)

image

oskarwrobel: I changed it because of
image
but looks like I changed too much :)

oskarwrobel: Fixed, not reverted but fixed :)

PR #208, opened 2017-11-24
t/201: Bulletproofed isDomNode helper when used in iframes. Removed isWindow logic.

### LINKED ISSUE #201: isDomNode() is confusing
Introduced in (without a ticket or at least a commit message): https://github.com/ckeditor/ckeditor5-utils/commit/fafd3d3.

It doesn't check if something is a node. It doesn't do much more than lodash's `isNative()` so I'd like to know why it was needed.

oleq:

Fix: Bulletproofed `isDomNode` helper when used in iframes. Removed `isWindow` logic from the helper. Closes #201.

### DIFF
--- a/src/dom/isdomnode.js
+++ b/src/dom/isdomnode.js
@@ -7,8 +7,6 @@
  * @module utils/dom/isdomnode
  */
 
-import isNative from '../lib/lodash/isNative';
-
 /**
  * Checks if the object is a native DOM Node.
  *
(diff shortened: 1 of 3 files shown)


src/dom/isdomnode.js
```diff
@@ -7,14 +7,20 @@
...
  * Checks if the object is a native DOM Node.
  *
  * @param {*} obj
  * @returns {Boolean}
  */
 export default function isDomNode( obj ) {
-	return !!( obj && isNative( obj.addEventListener ) );
+	if ( obj ) {
+		if ( obj.defaultView ) {
+			return obj instanceof obj.defaultView.Document;
```
Reinmar: Is document a node?

oleq: Yes

image

Reinmar: Sorry, I actually wanted to know whether this code is needed at all and whether this code is semantically correct. But now I see that both things are true.

Reinmar:
The existing `isWindow()` made me think about:

image

That may be some direction to consider if the one proposed by you in this PR will turn to be incorrect one day. Something like this should do the job:

```js
isDomNode( obj ) {
    return Object.prototype.toString.apply( obj ).startsWith( '[object HTML' )
}
```

PR #210, opened 2017-11-27
Don't throw error when trying to unbind attribute that is not bound.

### LINKED ISSUE #5: Observable#unbind: unbinding attribute which is not bound throws ambiguous error
Moved from https://github.com/ckeditor/ckeditor5-core/issues/153

``` js
A.bind( 'x' ).to( B, 'a' );
A.unbind( 'totallyNotBoundToAnythingOrEvenNotExistingInA' );
```

**Actual**: `TypeError: Cannot read property 'to' of undefined`

**Expected**: A meaningful error.

jodator:

Enhancement: Don't throw error when trying to unbind attribute that is not bound. Closes #5.

---

### Additional information

* Small fix I suppose.

### DIFF
--- a/tests/observablemixin.js
+++ b/tests/observablemixin.js
@@ -739,6 +739,14 @@ describe( 'Observable', () => {
 			observable.unbind();
 		} );
 
+		it( 'should not fail when unbinding non-binded property', () => {
+			const observable = new Observable();
+
+			observable.bind( 'foo' ).to( car, 'color' );
+
+			observable.unbind( 'bar' );
+		} );
+
 		it( 'should throw when non-string attribute is passed', () => {
 			expect( () => {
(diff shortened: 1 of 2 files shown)


tests/observablemixin.js
```diff
@@ -739,6 +739,14 @@ describe( 'Observable', () => {
 			observable.unbind();
 		} );
 
+		it( 'should not fail when unbinding non-binded property', () => {
+			const observable = new Observable();
+
+			observable.bind( 'foo' ).to( car, 'color' );
+
+			observable.unbind( 'bar' );
```
oleq: I'd use `expect( () => {} ).to.not.throw();` here to be consistent with other tests.

jodator: I also reword the test name.

PR #213, opened 2017-12-12
Fix exception that occur after dragging the selection through the image

ma2ciek:

Fix: Added fallback for `Rect.getDomRangeRects()` if the given range starts in text node. Closes ckeditor/ckeditor5-ui#317.

### DIFF
--- a/src/dom/rect.js
+++ b/src/dom/rect.js
@@ -7,6 +7,8 @@
  * @module utils/dom/rect
  */
 
+/* global Node */
+
 import isRange from './isrange';
 import isWindow from './iswindow';
 import isElement from '../lib/lodash/isElement';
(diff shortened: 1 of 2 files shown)


src/dom/rect.js
```diff
@@ -361,12 +363,19 @@ export default class Rect {
...
 		// instead and adjust rect's width to simulate the actual geometry of such range.
 		// https://github.com/ckeditor/ckeditor5-utils/issues/153
+		// https://github.com/ckeditor/ckeditor5-ui/issues/317
 		else {
-			const startContainerRect = new Rect( range.startContainer.getBoundingClientRect() );
-			startContainerRect.right = startContainerRect.left;
-			startContainerRect.width = 0;
+			let startContainer = range.startContainer;
+
+			if ( startContainer.nodeType === Node.TEXT_NODE ) {
```
Reinmar: Would be good to have an `isText()` util for that.

Reinmar: Actually, we have it in the DomConverter already :D And we use similar logic in `scroll.js`. So at least `scroll.js` and now `rect.js` should use it. DomConverter not necessarily.

ma2ciek: Can I make the above change as a follow-up?

Reinmar: 👍

PR #218, opened 2018-01-31
Implement isText() helper

### LINKED ISSUE #214: Implement the dom/isText helper
ATM we have `isWindow()`, `isDomNode()` and `isRange()`. The new helper should be used wherever possible (comment below) and work accross–iframes to avoid issues in the future.

A follow-up of https://github.com/ckeditor/ckeditor5-utils/pull/213#discussion_r156635229.

ma2ciek:

Other: Introduced `isText()` helper. Closes #214.

---

### Additional information
PRs in other packages:
- https://github.com/ckeditor/ckeditor5-engine/pull/1259

### DIFF
--- a/src/dom/istext.js
+++ b/src/dom/istext.js
@@ -1,0 +1,18 @@
+/**
+ * @license Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module utils/dom/istext
+ */
+
+/**
+ * Checks if the object is a native DOM Text node.
+ *
+ * @param {*} obj
+ * @returns {Boolean}
+ */
+export default function isText( obj ) {
+	return Object.prototype.toString.call( obj ) == '[object Text]';
+}
--- a/src/dom/rect.js
+++ b/src/dom/rect.js
@@ -7,13 +7,12 @@
  * @module utils/dom/rect
  */
 
-/* global Node */
-
 import isRange from './isrange';
 import isWindow from './iswindow';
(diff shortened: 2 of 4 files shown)


src/dom/istext.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+/**
+ * Checks if the object is a native DOM Text Node.
+ *
+ * @param {*} obj
+ * @returns {Boolean}
+ */
+export default function isText( obj ) {
+	return Object.prototype.toString.call( obj ) == '[object Text]';
```
ma2ciek: Not sure if it's better checking `nodeType` or calling `toString()`. IMO both options should be good here.

Reinmar: Checking `nodeType` seems to be a too naive duck typing. Calling `toString()` is slightly safer. Eventually, we could switch to `instanceof` but starting with `toString()` seems to be simpler.

ma2ciek: Simple `instanceof` check easily breaks in iframes, so it would be more complicated to implement it correctly.

oleq: BTW: I wonder why `isDomNode` instead of `isNode`? It's under `src/dom/`. If we go with `isText` we should go with `isNode` too.

Reinmar: I forgot to comment on this – yep, `isNode()` would be better. @ma2ciek could you report a separate ticket for it and do the change?

ma2ciek: Sure, I'll handle it.

ma2ciek: https://github.com/ckeditor/ckeditor5-utils/issues/219

PR #225, opened 2018-02-08
Feature: Introduce `bind().toMany()` binding chain in `ObservableMixin`.

### LINKED ISSUE #224: Introduce Observable#bind()#toMany()
Binding one property to many (reduce-like) is required by https://github.com/ckeditor/ckeditor5-ui/pull/361.

Reinmar:

Type: Feature: Introduce `bind().toMany()` binding chain in `ObservableMixin`. Closes #224.

### DIFF
--- a/src/observablemixin.js
+++ b/src/observablemixin.js
@@ -136,6 +136,7 @@ const ObservableMixin = {
 
 		// @typedef {Object} BindChain
 		// @property {Function} to See {@link ~ObservableMixin#_bindTo}.
+		// @property {Function} toMany See {@link ~ObservableMixin#_bindToMany}.
 		// @property {module:utils/observablemixin~Observable} _observable The observable which initializes the binding.
 		// @property {Array} _bindProperties Array of `_observable` properties to be bound.
 		// @property {Array} _to Array of `to()` observable–properties (`{ observable: toObservable, properties: ...toProperties }`).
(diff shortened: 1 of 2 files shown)


src/observablemixin.js
```diff
@@ -414,6 +416,40 @@ function bindTo( ...args ) {
...
+}
+
+// Returns an array of binding components for
+// {@link Observable#bind} from a set of iterable observables.
+//
+// @param {Iterable.<Observable>} observables
+// @param {String} attribute
+// @returns {Array.<String>}
+function getBindingTargets( observables, attribute ) {
+	return Array.prototype.concat( ...observables.map( observable => [ observable, attribute ] ) );
```
ma2ciek: Can't it be just that:
```js
return observables.map( observable => [ observable, attribute ] );
```

Reinmar: It seems so. I admit that I haven't even looked at the code ;)

Reinmar: Ah, no. It needs to be a `concat()` call. Because it takes `[1,2],[3,4],...` arrays and returns `[1,2,3,4...]`. Am I right, @jodator?

ma2ciek: Hmm, after short analyzing I've found that it might be correct as the result of this fn is `[ observable1, attribute, observable2, attribute ... ]`.

But docs are incorrect here anyway, `observables` can't be here iterable and the return type is different.

Reinmar: Feel free to fix :)

jodator: Yeah - return type should be `Array.<Observable|String>` as the result is as @ma2ciek wrote.

ma2ciek: Ok, I'll change it.

PR #226, opened 2018-02-08
Improve docs and implementation of `getBindingTargets` method

ma2ciek:

Internal: Improved docs and implementation of `getBindingTargets` method.

### DIFF
--- a/src/observablemixin.js
+++ b/src/observablemixin.js
@@ -419,7 +419,7 @@ function bindTo( ...args ) {
 // Binds to an attribute in a set of iterable observables.
 //
 // @private
-// @param {Iterable.<Observable>} observables
+// @param {Array.<Observable>} observables
 // @param {String} attribute
 // @param {Function} callback
 function bindToMany( observables, attribute, callback ) {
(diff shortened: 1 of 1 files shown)


src/observablemixin.js
```diff
@@ -445,9 +445,12 @@ function bindToMany( observables, attribute, callback ) {
 //
 // @param {Iterable.<Observable>} observables
 // @param {String} attribute
-// @returns {Array.<String>}
+// @returns {Array.<String|Observable>}
 function getBindingTargets( observables, attribute ) {
-	return Array.prototype.concat( ...observables.map( observable => [ observable, attribute ] ) );
+	const observableAndAttributePairs = Array.from( observables ).map( observable => [ observable, attribute ] );
```
Reinmar: Why do you change implementation? Is support for iterables needed?

ma2ciek: I can revert it and allow only arrays here.

But there's the second change I've made. I changed `Array.prototype.concat()` to `Array.prototype.concat.apply()`. It looked strange to me, that `this` refers to `undefined` here.

Reinmar: It's not `undefined` – it equalled `Array.prototype` ;) And yes – it was a bit ugly and you can change it.

PS. Please remember about updating the docs after removing `Array.from()` call.

Reinmar: PPS. There would be nothing wrong with supporting iterables if that was required by the public API. This is a private function so it should do only what it's needed from it.

ma2ciek: It reflects also on the `bindToMany` function and the exposed `bind().toMany()`. There's no need for supporting iterables in the exposed function I suppose, so I'll change it.

ma2ciek:
Fixed.

Reinmar:
BTW, the correct msg was:

> Internal: Improved docs and implementation of the `getBindingTargets()` function.

PR #227, opened 2018-02-15
Docs: Update utils/uid documentation.

### LINKED ISSUE #192: The `uid` function has docs describing old behavior
The `uid` method changed to be more UUID like but it's docs does not changed.

jodator:

Docs: Update utils/uid documentation. Closes #192.

---

### Additional information

* I cannot look longer on the documentation of this method... I've also updated the test to reflect current uid format.

### DIFF
--- a/src/uid.js
+++ b/src/uid.js
@@ -8,10 +8,10 @@
  */
 
 /**
- * Returns a unique id. This id is a number (starting from 1) which will never get repeated on successive calls
- * to this method.
+ * Returns a unique id. This id consist of an 'e' character and a randomly generated string of 32 aphanumeric characters.
+ * Each character in uid string represents a hexadecimal digit (base 16).
  *
- * @returns {String} A number representing the id.
+ * @returns {String} A hexadecimal number representing the id.
  */
 export default function uid() {
 	let uuid = 'e'; // Make sure that id does not start with number.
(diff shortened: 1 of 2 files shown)


jodator:
@Reinmar probably not an issue but there are some UUID implementations that conforms to the RFCs: https://github.com/broofa/node-uuid (have browser version).

Reinmar:
I think that we called it `uid()` instead of `uuid()` for a reason. We need a unique id and its format does not matter. In fact, the fact that it starts with `e` doesn't make any difference. These could be random characters and that's all.

Reinmar:
Actually, this is a question to @czerwonkabartosz. I think that you guys needed changes in this function, right? https://github.com/ckeditor/ckeditor5-utils/issues/192 says that this method's implementation has been changed, but when did it happen and why?

czerwonkabartosz:
Hmmm I'm pretty sure that we do not use that function in our code. I've checked our code and this function is not used by us.

Reinmar:
Thanks. I checked who touched this file and it was changed in https://github.com/ckeditor/ckeditor5-utils/commit/d7a9184fe0e40186b2f8238fd584afec175aeca6. So it wasn't following some requirements, AFAICS. https://github.com/ckeditor/ckeditor5-utils/issues/97 mentions that it was copied from CKE4 :D

Let's leave it like this.

czerwonkabartosz:
I've check changes in https://github.com/ckeditor/ckeditor5-utils/commit/d7a9184fe0e40186b2f8238fd584afec175aeca6 and now I understand... Id items from the collection are saved in our database and should be unique. 

But for us format does not matter ;) it should be unique and this is our requirement.

PR #232, opened 2018-03-27
Removed translation-utils's add()

Reinmar:

Other: Removed `translation-service~add()` function in favor of using safer, always available, manual approach to extending `CKEDITOR_TRANSLATIONS` global. See ckeditor/ckeditor5#914.

BREAKING CHANGES: The `translation-service~add()` function was removed. See `translation-utils` module docs to learn how to add translations.

### DIFF
--- a/src/translation-service.js
+++ b/src/translation-service.js
@@ -7,37 +7,37 @@
 
 /**
  * @module utils/translation-service
- */
-
-let dictionaries = {};
-
-/**
- * Adds package translations to existing ones.
- * These translations will later be available for {@link module:utils/translation-service~translate translate}.
  *
- *		add( 'pl', {
- *			'OK': 'OK',
- *			'Cancel [context: reject]': 'Anuluj'
- *		} );
- *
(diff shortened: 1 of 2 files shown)


src/translation-service.js
```diff
@@ -7,37 +7,37 @@
...
+ *			// Extend the dictionary for the given language.
+ *			Object.assign( dictionary, translations );
+ *		}
  */
-export function add( lang, translations ) {
-	dictionaries[ lang ] = dictionaries[ lang ] || {};
 
-	Object.assign( dictionaries[ lang ], translations );
+// Initialize CKEDITOR_TRANSLATIONS if it's not initialized.
+if ( !window.CKEDITOR_TRANSLATIONS ) {
```
Reinmar: Again, I think that such an access to the global variable is unsafe in envs like Electron, but let's go with it for now. I'll only change it so it imports the global object from `src/global.js`.

Reinmar: Or maybe I won't... because it's `src/dom/global.js`. I'll report a followup.

PR #238, opened 2018-04-16
Introduced fastDiff()

### LINKED ISSUE #235: Introduce fastDiff()
Extracted from https://github.com/ckeditor/ckeditor5-engine/issues/403 (see https://github.com/ckeditor/ckeditor5-engine/issues/403#issuecomment-380467350).

The whole idea behind `fastDiff` is to introduce a faster diffing function (as our [`diff`](https://github.com/ckeditor/ckeditor5-utils/blob/master/src/diff.js) [has some performance issues in specific cases](https://github.com/ckeditor/ckeditor5-engine/issues/403#issuecomment-375260327)) which will be able to diff two strings determining at what index was the first and last change and what text was changed.

The main purpose is to use it in the renderer while implementing smart text nodes updates (https://github.com/ckeditor/ckeditor5-engine/issues/403). So based on a `fastDiff` result we would like to use [`CharacterData.insertData`](https://developer.mozilla.org/en-US/docs/Web/API/CharacterData#Methods) and [`CharacterData.deleteData`](https://developer.mozilla.org/en-US/docs/Web/API/CharacterData#Methods) to smartly (partially) replace exisitng text node content when rendering new text.

f1ames:

Other: Introduced `fastDiff` diffing function. Closes #235.

---

### Additional information

See #235 for more details.

### DIFF
--- a/src/fastdiff.js
+++ b/src/fastdiff.js
@@ -1,0 +1,156 @@
+/**
+ * @license Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module utils/fastdiff
+ */
+
+/**
+ * Finds position of the first and last change in the given strings and generates set of changes. Set of changes
+ * can be applied to the input text in order to transform it into the output text, for example:
+ *
+ *		let input = '12abc3';
+ *		const output = '2ab';
+ *		const changes = fastDiff( input, output );
+ *
+ *		console.log( changes );
+ *		// [ { index: 0, type: 'insert', values: [ '2', 'a', 'b' ] }, { index: 3, type: 'delete', howMany: 6 } ]
+ *
+ *		changes.forEach( change => {
+ *			if ( change.type == 'insert' ) {
+ *				input = input.substring( 0, change.index ) + change.values.join( '' ) + input.substring( change.index );
+ *			} else if ( change.type == 'delete' ) {
+ *				input = input.substring( 0, change.index ) + input.substring( change.index + change.howMany );
+ *			}
+ *		} );
+ *
+ *		input == output; // -> true
+ *
+ * The output format of this function is compatible with {@link module:utils/difftochanges~diffToChanges} output format.
+ *
+ * @param {String} oldText Input string.
+ * @param {String} newText Input string.
+ * @returns {Array} Array of changes.
+ */
+export default function fastDiff( oldText, newText ) {
+	// Check if both texts are equal.
+	if ( oldText === newText ) {
+		return [];
+	}
+
+	const changeIndexes = findChangeBoundaryIndexes( oldText, newText );
+	return changeIndexesToChanges( newText, changeIndexes );
+}
+
+// Finds position of the first and last change in the given strings. For example:
+//
+//		const indexes = findChangeBoundaryIndexes( '1234', '13424' );
+//		console.log( indexes ); // { firstIndex: 1, lastIndexOld: 3, lastIndexNew: 4 }
+//
+// The above indexes means that in `oldText` modified part is `1[23]4` and in the `newText` it is `1[342]4`.
+// Based on such indexes, array with `insert`/`delete` operations which allows transforming
+// old text to the new one could be generated.
+//
+// @param {String} oldText
+// @param {String} newText
+// @returns {Object}
+// @returns {Number} return.firstIndex Position of the first change in both strings (always the same for both).
+// @returns {Number} result.lastIndexOld Position of the last change in `oldText` string.
+// @returns {Number} result.lastIndexNew Position of the last change in `newText` string.
+function findChangeBoundaryIndexes( oldText, newText ) {
+	const oldTextLength = oldText.length;
+	const newTextLength = newText.length;
+
+	// Find first change.
+	// Iterate over both strings starting from the beginning to find position of the first different character.
+	// If not found, it means first change is at the end of the string.
+	let firstIndex = oldTextLength;
+	for ( let i = 0; i < oldTextLength; i++ ) {
+		if ( i >= newTextLength || oldText[ i ] !== newText[ i ] ) {
+			firstIndex = i;
+			break;
+		}
+	}
+
+	// Find last change.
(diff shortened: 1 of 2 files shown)


src/fastdiff.js
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * @module utils/fastdiff
+ */
+
+/**
+ * Finds position of the first and last change in the given strings and generates set of changes. Set of changes
+ * can be applied to the input text in order to transform it into the output text, for example:
+ *
+ *		let input = '12abc3';
```
scofalik: I'd add at least two more, basic examples. One example: adding a few consecutive characters somewhere in the string. Second example: removing a few consecutive characters from the string. The example you provided is actually kind of an edge case, where characters were removed from the beginning and from the end. So, the output might actually be surprising for someone.

Other than that, I am not sure about the "usage" snippet. I have a feeling that the example is a bit uninspiring. Even though it shows how to use the output, somebody would think "why would I want to do it?". I have a feeling that we might want to drop this snippet.

I'd also shorten the examples:
```
* (...) for example:
*
*		fastDiff( '12a',	'12xyza' );	// [ { ... } ]
*		fastDiff( '12a',	'12aa' );	// [ { ... } ]
*		fastDiff( '12xyza',	'12a' );	// [ { ... } ]
*		fastDiff( '12aa',	'12a' );	// [ { ... } ]
```

scofalik: If the output is too long for the comment in the same line, it can be above or below starting with "Following produces:" and with extra line between each example.

src/fastdiff.js
```diff
@@ -0,0 +1,156 @@
...
+
+// Finds position of the first and last change in the given strings. For example:
+//
+//		const indexes = findChangeBoundaryIndexes( '1234', '13424' );
+//		console.log( indexes ); // { firstIndex: 1, lastIndexOld: 3, lastIndexNew: 4 }
+//
+// The above indexes means that in `oldText` modified part is `1[23]4` and in the `newText` it is `1[342]4`.
+// Based on such indexes, array with `insert`/`delete` operations which allows transforming
+// old text to the new one could be generated.
+//
```
scofalik: I'd add a note that it is assumed that the strings are different.

src/fastdiff.js
```diff
@@ -0,0 +1,156 @@
...
+function findChangeBoundaryIndexes( oldText, newText ) {
+	const oldTextLength = oldText.length;
+	const newTextLength = newText.length;
+
+	// Find first change.
+	// Iterate over both strings starting from the beginning to find position of the first different character.
+	// If not found, it means first change is at the end of the string.
+	let firstIndex = oldTextLength;
+	for ( let i = 0; i < oldTextLength; i++ ) {
+		if ( i >= newTextLength || oldText[ i ] !== newText[ i ] ) {
```
scofalik: `i >= newTextLength` should not be needed. If it is true, then `newText[ i ] == undefined` so surely the second condition will be met.

src/fastdiff.js
```diff
@@ -0,0 +1,156 @@
...
+	const maxLength = Math.max( oldTextReversedLength, newTextReversedLength );
+
+	let lastIndexOld;
+	let lastIndexNew;
+	for ( let i = 0; i < maxLength; i++ ) {
+		// Initial text -> after removing identical part -> reversed:
+		// oldText: '321ba' -> '21ba' -> 'ab12'
+		// newText: '31ba'  -> '1ba'  -> 'ab1'
+		// { firstIndex: 1, lastIndexOld: 2, lastIndexNew: 1 }
+		if ( i >= newTextReversedLength ) {
```
scofalik: Those additional conditions are probably not needed either.

scofalik:
I've spent some time analyzing this code, cause it looked to me that the `fastDiff` function is too complicated. There are many variables initialized, strings are reversed and cut and you quickly stop to understand what's what and why.

So, the algorithm itself cannot be simpler. Of course, finding the first change is easy. Finding the last change is more difficult. Cutting and reversing the string leads in fact to the simpler implementation (you could do the same using proper offsets but it would be even worse).

However, I think that it may be written in a simpler way. I'll propose something.

src/fastdiff.js
scofalik: Well, unfortunately, it is also needed.

scofalik:
BTW. I've also added two tests - one that I thought should fail for my algorithm, and the one that is used in the docs.

Reinmar:
> Since I am subjective on this matter, maybe @Reinmar could take a look and say which solution looks cleaner.

I'm sure you'll be able to figure out which one you like more, guys ;) You know far more about this code than I'll be able to learn now.

src/fastdiff.js
```diff
@@ -0,0 +1,153 @@
...
+// Based on such indexes, array with `insert`/`delete` operations which allows transforming
+// old text to the new one could be generated.
+//
+// It is expected that `oldText` and `newText` are different.
+//
+// @param {String} oldText
+// @param {String} newText
+// @returns {Object}
+// @returns {Number} return.firstIndex Index of the first change in both strings (always the same for both).
+// @returns {Number} result.lastIndexOld Index of the last common character in `oldText` string looking from back.
```
f1ames: I think it should be:

> result.lastIndexOld Index of the last common character in `oldText` string.

or

> result.lastIndexOld Index of the first common character in `oldText` string looking from back.

Because looking for the last common character starting from the back, means you are looking for the first common character basically.

f1ames:
I just added some docs improvements. TBH @scofalik, your refactor made it more readable and simplified it a little so I'm fine with it. Looks good IMHO.

> Other than that, I am not sure about the "usage" snippet. I have a feeling that the example is a bit uninspiring. Even though it shows how to use the output, somebody would think "why would I want to do it?". I have a feeling that we might want to drop this snippet.

I see you left the snippet, was it on purpose? I wasn't sure either, but it gives some general picture how `fastDiff` output can be used so maybe we could left it there.

Merged latest changes from `master` too.

I think it is ready for review, from my perspective it looks good, nothing to add here.

f1ames:
As agreed with @scofalik F2F, I have merged the PR as it was ready 🎉

PR #240, opened 2018-05-09
Fix: parentNode error.

tiaguinho:
This validations fix the problem with "Enter" inside an iframe.
I have this issue using ckeditor inside an Angular project.

I think this PR also fix [#930](https://github.com/ckeditor/ckeditor5/issues/930)

## Adicional information

Because the main project isn't an Angular project,  my project (let's call FOO) have to live inside an iframe.
To make sure that our preview isn't affected by FOO style, the preview also live inside an iframe. So the structure is like this:
```typescript
MAIN PROJECT:
   - FOO iframe:
      - PREVIEW iframe (ckeditor live's here)
```

When we press enter, backspace or delete, the error ```Uncaught TypeError: Cannot read property 'parentNode' of null``` is throw.

I have tested this PR proposal inside FOO and the error was gone.

### DIFF
--- a/src/dom/scroll.js
+++ b/src/dom/scroll.js
@@ -45,24 +45,24 @@ export function scrollViewportToShowTarget( { target, viewportOffset = 0 } ) {
 			firstAncestorToScroll = getParentElement( currentFrame );
 		}
 
-		if (firstAncestorToScroll) {
+		if ( firstAncestorToScroll ) {
 			// Scroll the target's ancestors first. Once done, scrolling the viewport is easy.
-			scrollAncestorsToShowRect(firstAncestorToScroll, () => {
+			scrollAncestorsToShowRect( firstAncestorToScroll, () => {
 				// Note: If the target does not belong to the current window **directly**,
 				// i.e. it resides in an iframe belonging to the window, obtain the target's rect
 				// in the coordinates of the current window. By default, a Rect returns geometry
 				// relative to the current window's viewport. To make it work in a parent window,
 				// it must be shifted.
-				return getRectRelativeToWindow(target, currentWindow);
-			});
+				return getRectRelativeToWindow( target, currentWindow );
+			} );
 
 			// Obtain the rect of the target after it has been scrolled within its ancestors.
 			// It's time to scroll the viewport.
-			const targetRect = getRectRelativeToWindow(target, currentWindow);
+			const targetRect = getRectRelativeToWindow( target, currentWindow );
 
-			scrollWindowToShowRect(currentWindow, targetRect, viewportOffset);
+			scrollWindowToShowRect( currentWindow, targetRect, viewportOffset );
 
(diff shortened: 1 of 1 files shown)


tiaguinho:
@Reinmar can you take a look?

Reinmar:
Hi! Thanks for the PR. I'm leaving it for @oleq to review, because he knows this code better. 

However, to merge this change we need to have an automated test verifying the patch and the code coverage back to 100%. Ideally, 

You can read about our testing env in https://docs.ckeditor.com/ckeditor5/latest/framework/guides/contributing/testing-environment.html.

src/dom/scroll.js
```diff
@@ -45,29 +45,33 @@ export function scrollViewportToShowTarget( { target, viewportOffset = 0 } ) {
...
-		scrollWindowToShowRect( currentWindow, targetRect, viewportOffset );
-
-		if ( currentWindow.parent != currentWindow ) {
-			// Keep the reference to the <iframe> element the "previous current window" was
-			// rendered within. It will be useful to re–calculate the rect of the target
-			// in the parent window's relative geometry. The target's rect must be shifted
-			// by it's iframe's position.
-			currentFrame = currentWindow.frameElement;
-			currentWindow = currentWindow.parent;
+		if ( firstAncestorToScroll ) {
```
oleq: What does this condition actually do? In which cases there's no `firstAncestorToScroll`? 

As I [wrote](https://github.com/ckeditor/ckeditor5/issues/930#issuecomment-377495214) I tested the editor inside an iframe (not JSFiddle), and I couldn't reproduce the issue. What makes your use–case different?

This condition disables scrolling (ancestors, window) completely, which is wrong IMO. I see no good reason why it should be disabled; the bugs should be fixed instead.

tiaguinho: When I start debugging ckeditor inside my application, I get the following result when I press ```enter```:

* ```currentFrame``` equal null;
* ```currentWindow``` different of ```targetWindow```;
* ```getParentElement``` is called with a null param;
* console output error when the ```getParentElement``` try to return ***parentNode*** of null.

The changes I have made just make sure that ```elementOrRange``` isn't null.

tiaguinho:
@oleq I don't know if the problem is related to the fact that my Angular app is running inside an iframe, but I rebuild ckeditor using my fork and the error was fixed.

Reinmar:
@oleq, I've just encountered https://github.com/ckeditor/ckeditor5/issues/930 and we should investigate this.

oleq:
@tiaguinho I found the root of the issue and moved the code to a different PR because I was unable to contribute to this one. Can you confirm that it works for you in https://github.com/ckeditor/ckeditor5-utils/pull/242?

Reinmar:
I merged https://github.com/ckeditor/ckeditor5-utils/pull/242 already, so if you'll find that the issue still occurs in your case, please report a new ticket.

PR #241, opened 2018-05-28
Feature: Added `beforeChange` event to `ObservableMixin`.

### LINKED ISSUE #171: What is the best way to block observable propertie's value?
This is a common practice e.g. when one feature wants to control a state of another feature's command.

See https://github.com/ckeditor/ckeditor5-core/blob/4d768dff3facd329b881859a4b831d2dd6a3e3f4/tests/command.js#L118-L135

It overrides some property and then ensures that no one changes it back.

In that test I used a high priority listener and stopped the event so the change of the property value is not visible to the normal listeners. 

However, this breaks bindings because they rely on normal priority listeners.

So another option is to listen with a low priority listener and to not stop the event. This will fix the bindings but it means that the change event is visible twice. It's hard to predict implications of that. Infinite loops or blinking UI are possible.

So, what's the best way to achieve:

* minimal noise,
* working bindings,
* safety?

oskarwrobel:

Feature: Added `beforeChange` event to `ObservableMixin`. Closes #171.

---

### DIFF
--- a/src/observablemixin.js
+++ b/src/observablemixin.js
@@ -75,8 +75,17 @@ const ObservableMixin = {
 				// Allow undefined as an initial value like A.define( 'x', undefined ) (#132).
 				// Note: When properties map has no such own property, then its value is undefined.
 				if ( oldValue !== value || !properties.has( name ) ) {
-					properties.set( name, value );
-					this.fire( 'change:' + name, name, value, oldValue );
+					// Fire `beforeChange` event before the new value will be changed to make it possible
+					// to override observable property without affecting `change` event.
+					// See https://github.com/ckeditor/ckeditor5-utils/issues/171.
+					let newValue = this.fire( 'beforeChange:' + name, name, value, oldValue );
+
+					if ( newValue === undefined ) {
+						newValue = value;
+					}
+
+					properties.set( name, newValue );
+					this.fire( 'change:' + name, name, newValue, oldValue );
 				}
 			}
 		} );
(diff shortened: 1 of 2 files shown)


src/observablemixin.js
```diff
@@ -674,6 +683,37 @@ function attachBindToListeners( observable, toBindings ) {
  * @param {*} oldValue The previous property value.
  */
 
+/**
+ * Fired when a property value is going to be changed but is not changed yet
+ * (before the {@link module:utils/observablemixin~ObservableMixin#change} event is fired).
```
oskarwrobel: Incorrect link, unfortunately `module:utils/observablemixin~Observable#change` is incorrect as well.

oleq:
Why exactly do we need this change?

oskarwrobel:
It's all explained in the related ticket https://github.com/ckeditor/ckeditor5-utils/issues/171.

We need to have possibility to override observable without breaking the bindings.

oskarwrobel:
E.g.

```js
observableA.bind( 'state' ).to( observableB ); 

observableB.state = 'active';
console.log( observableA.state ) // 'active'

observableB.on( 'change:state', evt => {
  evt.stop();
  observableB.state = 'inactive';
}, { priority: 'highest' } );

observableB.state = 'pending';
console.log( observableA.state ) // 'active' but should be `inactive`
```
When observable change event is stopped in high priority listener then is not propagated farther so normal priority listeners (bindings) don't react on changes.

src/observablemixin.js
```diff
@@ -674,6 +683,36 @@ function attachBindToListeners( observable, toBindings ) {
  * @param {*} oldValue The previous property value.
  */
 
+/**
+ * Fired when a property value is going to be changed but is not changed yet (before the `change` event is fired).
+ *
+ * When some value will be set as {@link module:utils/eventinfo~EventInfo#return event return value} then
```
Reinmar: "as an event's return value"

src/observablemixin.js
```diff
@@ -674,6 +683,36 @@ function attachBindToListeners( observable, toBindings ) {
...
+ * this value become an observalbe new value.
+ *
+ *		observable.set( 'prop', 1 );
```
Reinmar: ❤️ very nice example; says more than 1000 words.

Reinmar:
I left one comment regarding wording. And besides that this PR is great.

Reinmar:
Actually, I'll change it and merge this PR.

PR #242, opened 2018-05-29
t/ckeditor5/930: Error should not be thrown when scrolling the viewport from within an iframe in a different domain

oleq:

Fix: Error should not be thrown when scrolling the viewport from within an iframe in a different domain. Closes ckeditor/ckeditor5#930.

---

### Additional information

It is a follow-up of https://github.com/ckeditor/ckeditor5-utils/pull/240/files because I was unable to push to the original PR.  I also simplified the solution and provided a valid test.

As for the cause of the issue, the error is thrown when the child and the parent windows have different domain/ports. In such situation, [for security reasons, `frameElement` is unavailable](https://developer.mozilla.org/en-US/docs/Web/API/Window/frameElement).

If worked for me in my sandbox because I used the same server. JSFiddle uses a different domain for live snippets and hence the error.

### DIFF
--- a/src/dom/scroll.js
+++ b/src/dom/scroll.js
@@ -45,24 +45,24 @@ export function scrollViewportToShowTarget( { target, viewportOffset = 0 } ) {
 			firstAncestorToScroll = getParentElement( currentFrame );
 		}
 
-		if (firstAncestorToScroll) {
+		if ( firstAncestorToScroll ) {
 			// Scroll the target's ancestors first. Once done, scrolling the viewport is easy.
-			scrollAncestorsToShowRect(firstAncestorToScroll, () => {
+			scrollAncestorsToShowRect( firstAncestorToScroll, () => {
 				// Note: If the target does not belong to the current window **directly**,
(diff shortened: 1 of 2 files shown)


src/dom/scroll.js
```diff
@@ -68,6 +68,14 @@ export function scrollViewportToShowTarget( { target, viewportOffset = 0 } ) {
 			// by it's iframe's position.
 			currentFrame = currentWindow.frameElement;
 			currentWindow = currentWindow.parent;
+
+			// If the current window has some parent but frameElement is inaccessible, then they have
+			// different domains/ports and, due to security reasons, accessing and scrolling
+			// the parent window won't be possible.
+			// See https://github.com/ckeditor/ckeditor5/issues/930.
+			if ( !currentFrame ) {
+				return;
```
Reinmar: One of the cases where `log.debug()` could be a big help :)

cc @pjasiun

pjasiun: Added to my list :)

PR #248, opened 2018-07-05
Feature: Add env.isEdge property.

jodator:

Feature: Introduce `env.isEdge` property.

---

### Additional information

* Required for:  https://github.com/ckeditor/ckeditor5-widget/pull/46

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -53,5 +53,5 @@ export function isMac( userAgent ) {
  * @returns {Boolean} Whether User Agent is Edge or not.
  */
 export function isEdge( userAgent ) {
-	return userAgent.indexOf( 'edge' ) > -1;
+	return !!userAgent.match( /edge\/(\d+.?\d*)/ );
 }
--- a/src/env.js
+++ b/src/env.js
@@ -23,7 +23,15 @@ const env = {
 	 * @static
 	 * @member {Boolean} module:utils/env~env#isMac
 	 */
-	isMac: isMac( userAgent )
+	isMac: isMac( userAgent ),
+
+	/**
+	 * Indicates that application is running on Edge browser.
+	 *
+	 * @static
+	 * @member {Boolean} module:utils/env~env#isEdge
+	 */
+	isEdge: isEdge( userAgent )
 };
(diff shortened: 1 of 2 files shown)


src/env.js
```diff
@@ -37,3 +45,13 @@ export default env;
...
 }
+
+/**
+ * Checks if User Agent represented by the string is Edge browser.
+ *
+ * @param {String} userAgent **Lowercase** `navigator.userAgent` string.
+ * @returns {Boolean} Whether User Agent is Edge or not.
+ */
+export function isEdge( userAgent ) {
+	return userAgent.indexOf( 'edge' ) > -1;
```
Reinmar: That's unsafe. There are many words with "edge" in them. Something more precise would be better.

Reinmar: BTW, feel free to look into CKE4's code:

https://github.com/ckeditor/ckeditor-dev/blob/master/core/env.js#L20

jodator: @Reinmar I'm not 100% sure but it looks like `[ /]` in that regex could be restricted to:

```js
/edge\/(\d+.?\d*)/
```

I couldn't find why there is space allowed...

jodator: ps.: https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/compatibility/hh869301(v=vs.85)

Describes only `Edge/<EdgeHTML Rev>.<Windows Build>` as possible value so I've set the regex accordingly to this.

tests/env.js
```diff
@@ -31,4 +37,19 @@ describe( 'Env', () => {
 			expect( isMac( 'foo' ) ).to.be.false;
 		} );
 	} );
+
+	describe( 'isEdge()', () => {
+		it( 'returns true for Edge UA strings', () => {
```
Reinmar: Not a single real UA string ;)

jodator: Yeah - I've copied from other test. I'm adding some real ones right now ;)

jodator: OK... I was too slow :disappointed:

PR #250, opened 2018-07-11
Made FocusTracker asynchronous

oskarwrobel:
### Suggested merge commit message ([convention](https://gith.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Fix: Made FocusTracker asynchronous. Closes https://github.com/ckeditor/ckeditor5-ui/issues/431.

---

### Additional information

### DIFF
--- a/src/focustracker.js
+++ b/src/focustracker.js
@@ -3,6 +3,8 @@
  * For licensing, see LICENSE.md.
  */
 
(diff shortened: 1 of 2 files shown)


pjasiun:
We decided to skip tests because automated tests does not work well with focus, and manual tests for this would be overkill.

PR #256, opened 2018-11-07
[Docs] Fix some typos

denisname:
Examples for fixes can be found in `tests/`.

There are some typos I don't know how to fix:

- **KeystrokeInfo**: the properties `altKey`, `ctrlKey` and `shiftKey` are all optional.
- **Options**: the properties `fitInViewport` and `limiter` are all optional.
- **Emitter#delegate**: type for `events` is `...String`.
- **Observable#bind**: type for `bindProperties` is `...String`.
- **Observable#unbind**: type for `unbindProperties` is `...String`.

### DIFF
--- a/src/dom/createelement.js
+++ b/src/dom/createelement.js
@@ -20,8 +20,8 @@ import { isString } from 'lodash-es';
  *
  * @param {Document} doc Document used to create element.
  * @param {String} name Name of the element.
- * @param {Object} attributes Object keys will become attributes keys and object values will became attributes values.
- * @param {Node|String|Array.<Node|String>} children Child or array of children. Strings will be automatically turned
+ * @param {Object|null} [attributes] Object keys will become attributes keys and object values will became attributes values.
+ * @param {Node|String|Array.<Node|String>} [children] Child or array of children. Strings will be automatically turned
  * into Text nodes.
  * @returns {Element} Created element.
  */
--- a/src/dom/getpositionedancestor.js
+++ b/src/dom/getpositionedancestor.js
@@ -12,7 +12,7 @@ import global from './global';
 /**
  * For a given element, returns the nearest ancestor element which CSS position is not "static".
  *
- * @param {HTMLElement} element Native DOM element to be checked.
+ * @param {HTMLElement} [element] Native DOM element to be checked.
  * @returns {HTMLElement|null}
  */
 export default function getPositionedAncestor( element ) {
--- a/src/keyboard.js
+++ b/src/keyboard.js
@@ -43,7 +43,7 @@ export const keyCodes = generateKnownKeyCodes();
  *
  * Note: Key names are matched with {@link module:utils/keyboard~keyCodes} in a case-insensitive way.
  *
- * @param {String|module:utils/keyboard~KeystrokeInfo} Key name (see {@link module:utils/keyboard~keyCodes})
+ * @param {String|module:utils/keyboard~KeystrokeInfo} key name (see {@link module:utils/keyboard~keyCodes})
  * or a keystroke data object.
  * @returns {Number} Key or keystroke code.
  */
--- a/src/keystrokehandler.js
+++ b/src/keystrokehandler.js
@@ -81,7 +81,7 @@ export default class KeystrokeHandler {
 	 * @param {String|Array.<String|Number>} keystroke Keystroke defined in a format accepted by
 	 * the {@link module:utils/keyboard~parseKeystroke} function.
 	 * @param {Function} callback A function called with the
-	 * {@link module:engine/view/observer/keyobserver~KeyEventData key event data} object and
+	 * {@link module:utils/keyboard~KeystrokeInfo keystroke info} object and
 	 * a helper to both `preventDefault` and `stopPropagation` of the event.
 	 * @param {Object} [options={}] Additional options.
 	 * @param {module:utils/priorities~PriorityString|Number} [options.priority='normal'] The priority of the keystroke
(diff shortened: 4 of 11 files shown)


denisname:
Do you really need a _Contributor License Agreement_ for such trival _non code_ fixes?

Reinmar:
Thank you so much for tracking all those mistakes ❤️❤️❤️ 

> Do you really need a _Contributor License Agreement_ for such trival _non code_ fixes?

Frankly speaking, I don't know. But then, I'd go with the safe: yes. It just saves us from thinking about possible problems that could arise in all possible universes in the future.

Reinmar:
Hi @denisname! We received your CLA (along with 2 other PRs :)). I understand now, though, why you asked whether you have to sign it – if I had to do that via traditional mail, I'd be a bit surprised too (you could have signed it online). We'd like to thank you for your troubles with a small souvenir – could email me (p.koszulinski@cksource.com) your t-shirt size and address?

src/dom/createelement.js
```diff
@@ -20,8 +20,8 @@ import { isString } from 'lodash-es';
  *
  * @param {Document} doc Document used to create element.
  * @param {String} name Name of the element.
- * @param {Object} attributes Object keys will become attributes keys and object values will became attributes values.
- * @param {Node|String|Array.<Node|String>} children Child or array of children. Strings will be automatically turned
+ * @param {Object|null} [attributes] Object keys will become attributes keys and object values will became attributes values.
```
Reinmar: The `null` is unnecessary, since the param is marked as optional.

src/dom/getpositionedancestor.js
```diff
@@ -12,7 +12,7 @@ import global from './global';
 /**
  * For a given element, returns the nearest ancestor element which CSS position is not "static".
  *
- * @param {HTMLElement} element Native DOM element to be checked.
+ * @param {HTMLElement} [element] Native DOM element to be checked.
```
Reinmar: What would this function be doing if the element was not specified? This param is required for it do work.

src/keyboard.js
```diff
@@ -43,7 +43,7 @@ export const keyCodes = generateKnownKeyCodes();
  *
  * Note: Key names are matched with {@link module:utils/keyboard~keyCodes} in a case-insensitive way.
  *
- * @param {String|module:utils/keyboard~KeystrokeInfo} Key name (see {@link module:utils/keyboard~keyCodes})
+ * @param {String|module:utils/keyboard~KeystrokeInfo} key name (see {@link module:utils/keyboard~keyCodes})
```
Reinmar: That's a sentence so it should start with an uppercased letter.

src/keystrokehandler.js
```diff
@@ -81,7 +81,7 @@ export default class KeystrokeHandler {
 	 * @param {String|Array.<String|Number>} keystroke Keystroke defined in a format accepted by
 	 * the {@link module:utils/keyboard~parseKeystroke} function.
 	 * @param {Function} callback A function called with the
-	 * {@link module:engine/view/observer/keyobserver~KeyEventData key event data} object and
+	 * {@link module:utils/keyboard~KeystrokeInfo keystroke info} object and
```
Reinmar: That's a bit messy now, but it's actually called with `KeyEventData`. The `KeyEventData` objects implements `KeystrokeInfo` but it adds `preventDefault()` and `stopPropagation()` methods (which are mentioned later in this sentence. It'd be good to not rely on engine's object type, so that's why I said it's messy. But other than that, `KeyEventData` is more correct.

Reinmar:
I merged the PR. Some changes weren't correct (I left comments), but most of them were. Thank you!

src/dom/getpositionedancestor.js
denisname: It returns `null`. There is a [test](https://github.com/ckeditor/ckeditor5-utils/blob/7e6bd6631f8bcf677311736e7de095dd95327261/tests/dom/getpositionedancestor.js#L19-L21) for it. That said, _required_ make more sense indeed.

src/keyboard.js
denisname: When looking in documentation (https://ckeditor.com/docs/ckeditor5/latest/api/module_utils_keyboard.html#static-function-getCode), there is

```
getCode( Key ) → Number
```

Although all functions always have they parameters with lower case. Also in the _Parameters_ part, sentence is broken between "key" and "name".

What about the following code instead (twice key)?

```
* @param {String|module:utils/keyboard~KeystrokeInfo} key Key name (see {@link module:utils/keyboard~keyCodes})
```

By the way, how do you generate the documentation? I did not find anything in the doc about it. Do you use internal tools?

ma2ciek: Hi, @denisname!

> By the way, how do you generate the documentation? I did not find anything in the doc about it. Do you use internal tools?

Yes, we use the `JSDoc3` package with a few custom plugins available here - https://github.com/ckeditor/ckeditor5-dev/tree/master/packages/jsdoc-plugins to generate proper AST tree and then we generate the documentation using our internal tool called [`Umberto`](https://www.npmjs.com/package/umberto).

> What about the following code instead (twice key)?

> `* @param {String|module:utils/keyboard~KeystrokeInfo} key Key name (see {@link module:utils/keyboard~keyCodes})`

Yep, you're right, the sentence should include both `keys` as the first key must match the param's name.

Btw I've proposed https://github.com/ckeditor/ckeditor5/issues/1415, it should solve many of such typos and small documentation errors.

PR #265, opened 2019-01-03
Leave DOM Elements in configuration object returned from editor.config.get().

### LINKED ISSUE #264: DOM nodes passed to the `editor.config` shouldn't be cloned
When some DOM nodes are passed to the editor's config, each `editor.config.get()` call returns cloned ones, so it's impossible to access the real nodes. They should stay as they are.

Optionally we could use the callbacks to wrap these nodes as functions won't be cloned, but the first option would be better IMO.

jodator:

Other: DOM Elements will not be cloned when returned from config.get. Closes #264.

---

### Additional information

* @oskarwrobel: please check if the solution works for you.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -7,7 +7,7 @@
  * @module utils/config
  */
 
-import { isPlainObject, cloneDeep } from 'lodash-es';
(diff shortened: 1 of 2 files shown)


oskarwrobel:
Works 👌

PR #267, opened 2019-01-03
266: Introduce `Collection.has()` method.

### LINKED ISSUE #266: Introduce collection.has
I know you can use:

```js
collection.getIndex( item ) > 0
```
Or:
```js
!!collection.get( item )
```

But nothing is as pretty as:
```js
if( collection.has( item ) ) {
    //...
}
```

pjasiun:

Feature: introduce `Collection.has()` method. Closes #266.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -210,24 +210,32 @@ export default class Collection {
 	}
 
 	/**
-	 * Returns a boolean indicating whether the collection contains an item with the specified id or index.
+	 * Returns a boolean indicating whether the collection contains an item. Item id or index can be used instead of the
+	 * item.
 	 *
-	 * @param {String|Number} idOrIndex The item id or index in the collection.
+	 * @param {Object|String|Number} itemOrIdOrIndex The item, item id or item index in the collection.
 	 * @returns {Boolean} `true` if the collection contains the item, `false` otherwise.
 	 */
-	has( idOrIndex ) {
-		if ( typeof idOrIndex == 'string' ) {
-			return this._itemMap.has( idOrIndex );
-		} else if ( typeof idOrIndex == 'number' ) {
-			return !!this._items[ idOrIndex ];
-		}
+	has( itemOrIdOrIndex ) {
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -209,6 +209,29 @@ export default class Collection {
...
 	}
 
+	/**
+	 * Returns a boolean indicating whether the collection contains an item with the specified id or index.
+	 *
+	 * @param {String|Number} idOrIndex The item id or index in the collection.
+	 * @returns {Boolean} `true` if the collection contains the item, `false` otherwise.
+	 */
+	has( idOrIndex ) {
+		let item;
```
jodator: unused :P

oskarwrobel: Unused.

pjasiun: Outdated. Too late.

src/collection.js
```diff
@@ -209,6 +209,29 @@ export default class Collection {
...
+
+		if ( typeof idOrIndex == 'string' ) {
```
jodator: What if I want to check if `collection.has( item )` - looks handy also.

src/collection.js
```diff
@@ -209,6 +209,29 @@ export default class Collection {
...
+			return this._itemMap.has( idOrIndex );
+		} else if ( typeof idOrIndex == 'number' ) {
```
jodator: I'm not sure if checking index is a way to go. I'd expect something similar to `[].includes()` with additions for our needs.

src/collection.js
jodator: Oh, I see :disappointed: you fixed it already.

pjasiun:
> 2\. Apart of that I see that `collection.has( id )` might be useful but I'm not convinced about `collecion.has( index )` since AFAICS the collection keep track of proper index on adding item so one can check if `index > collection.length` to check if that index can be used. Also the semantics of `collection.has( index )` looks odd.

Agree. It's my bad, I was sure that all methods in this class which support `id`, handle also `index`, but then I realized that it is not true. Fixed.

PR #271, opened 2019-01-23
Replaced `for..of` with `Array.prototype.forEach`

pomek:

Other: Replaced `for..of` statement in `EventEmitter` with `Array.prototype.forEach`. This changes allows building a React application using `create-react-app@2`. Closes ckeditor/ckeditor5-react#40.


ma2ciek:
I've checked it and it indeed fixes the mentioned problem.

Issue #284, opened 2019-04-30
Observable Collection#length property.


oskarwrobel:
I need to display in UI a total number of items in a collection (it's not the first time). Unfortunately, collection `#length` property is not observable (it's a getter) so I can't simply bind it with UI. I need to create an observable property and listen to the collection `add` and `remove` events and update the observable.

I propose to change `Collection#length` to observable.

My only concern is that collection is used many times across the editor and this will increase the number of callbacks, so maybe it's not worth to create such a generic solution to cover only a few cases?

WDYT?

oskarwrobel:
> this will increase the number of callbacks

It depends on implementation. My first idea was to listen to the `add` and `remove` event and update observable, but it can be done in the other way. Just increment/decrement observable at the end of add and remove methods.

pjasiun:
I think it is a good idea. It should not be any performance problem, should be backward compatible and it might be useful in many cases.

oskarwrobel:
https://github.com/ckeditor/ckeditor5-utils/pull/285#issuecomment-489654713

PR #285, opened 2019-05-06
Introduced observable Collection#length property.

oskarwrobel:

Feature: Introduced observable Collection#length property. Closes https://github.com/ckeditor/ckeditor5-utils/issues/284.

---

### Additional information

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -33,6 +33,14 @@ export default class Collection {
 	 */
 	constructor( options = {} ) {
 		/**
+		 * The number of items available in the collection.
+		 *
+		 * @observable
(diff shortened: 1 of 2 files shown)


pjasiun:
More I think about it more I think there is no proper order to fire `change:length` and `add` (or `remove`) events.

If we fire `change:length` first, it means that it will be fired before bound collections are updated. It might be a reason of issues.

If we fire `add`/`remove` first, then there is be still old `length` when `add`/`remove` event is fired what will be terrible.

This is why, I think that having 2 separate events informing about basically the same action, is not the best idea and I am against introdusing length as observable.

PR #287, opened 2019-05-06
Fixed API tags for the env#isGecko property

### LINKED ISSUE #286: The env#isGecko property is not listed in the API docs
## Is this a bug report or feature request?

Other

## 💻 Version of CKEditor

CKEditor v5 @ 12.1.0

## 📋 Steps to reproduce

1. Open [`@ckeditor/ckeditor5-utils/src/env` namespace API docs](https://ckeditor.com/docs/ckeditor5/12.1.0/api/module_utils_env-env.html) (permalink to the cur version).
1. Locate `isGecko` property.

## ✅ Expected result

No property 😱

## 📃 Other details that might be useful

[Currently it's overwriting](https://github.com/ckeditor/ckeditor5-utils/blob/c274c22/src/env.js#L40) the `isEdge`, so `isEdge` has a wrong desc.

mlewand:

Docs: Fixed tags for the `env#isGecko` property so that it is listed in the API docs. Closes #286.

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -37,7 +37,7 @@ const env = {
 	 * Indicates that the application is running in Firefox (Gecko).
 	 *
 	 * @static
-	 * @member {Boolean} module:utils/env~env#isEdge
+	 * @member {Boolean} module:utils/env~env#isGecko
 	 */
 	isGecko: isGecko( userAgent ),
 
--- a/src/env.js
+++ b/src/env.js
@@ -21,7 +21,7 @@ const env = {
 	 * Indicates that the application is running on Macintosh.
 	 *
 	 * @static
-	 * @member {Boolean} module:utils/env~env#isMac
+	 * @type {Boolean}
 	 */
 	isMac: isMac( userAgent ),
 
@@ -29,7 +29,7 @@ const env = {
 	 * Indicates that the application is running in Microsoft Edge.
 	 *
 	 * @static
-	 * @member {Boolean} module:utils/env~env#isEdge
+	 * @type {Boolean}
 	 */
 	isEdge: isEdge( userAgent ),
 
@@ -37,7 +37,7 @@ const env = {
 	 * Indicates that the application is running in Firefox (Gecko).
 	 *
 	 * @static
-	 * @member {Boolean} module:utils/env~env#isGecko
+	 * @type {Boolean}
 	 */
 	isGecko: isGecko( userAgent ),
 
@@ -45,7 +45,7 @@ const env = {
 	 * Indicates that the application is running in Safari.
 	 *
 	 * @static
-	 * @member {Boolean} module:utils/env~env#isSafari
+	 * @type {Boolean}
 	 */
 	isSafari: isSafari( userAgent )
 };


src/env.js
```diff
@@ -37,7 +37,7 @@ const env = {
 	 * Indicates that the application is running in Firefox (Gecko).
 	 *
 	 * @static
-	 * @member {Boolean} module:utils/env~env#isEdge
+	 * @member {Boolean} module:utils/env~env#isGecko
```
ma2ciek: I guess that we can just use `@type {Boolean}` across this file.

mlewand: Good point, I'll check that tomorrow.

ma2ciek: And actually, something weird happens now during JSDoc generation:

```json
{
	"comment": "/**\n\t * Indicates that the application is running in Microsoft Edge.\n\t *\n\t * @static\n\t * @member {Boolean} module:utils/env~env#isEdge\n\t */",
	"meta": {
		"range": [
			513,
			652
		],
		"filename": "env.js",
		"lineno": 28,
		"path": "/Users/maciejbukowski/workspace/cloud-features/packages/ckeditor5/packages/ckeditor5-utils/src",
		"code": {}
	},
	"description": "<p>Indicates that the application is running in Microsoft Edge.</p>",
	"scope": "instance",
	"kind": "member",
	"name": "isEdge",
	"type": {
		"names": [
			"Boolean"
		]
	},
	"memberof": "module:utils/env~env",
	"longname": "module:utils/env~env#isEdge"
},
{
	"comment": "",
	"meta": {
		"range": [
			654,
			681
		],
		"filename": "env.js",
		"lineno": 34,
		"path": "/Users/maciejbukowski/workspace/cloud-features/packages/ckeditor5/packages/ckeditor5-utils/src",
		"code": {
			"id": "astnode100038570",
			"name": "isEdge",
			"type": "CallExpression",
			"value": ""
		}
	},
	"undocumented": true,
	"name": "isEdge",
	"longname": "module:utils/env~env.isEdge",
	"kind": "member",
	"memberof": "module:utils/env~env",
	"scope": "static"
},
```

I guess that the second is ignored.
Maybe the output will be correctly generated with just the following:

```
/**
 * Indicates that the application is running in Firefox (Gecko).
 *
 * @type {Boolean}
 */
isGecko: isGecko( userAgent )
```

mlewand: @ma2ciek I see a lot of entries of this kind (with `undocumented: true`) throughout `output.json`.

Nonetheless I see that using shorter notation merged the two entries into one and the properties are shown in Umberto as static (and [this is missing in the current docs version](https://ckeditor.com/docs/ckeditor5/12.1.0/api/module_utils_env-env.html)).

ma2ciek: Hm. I'm not sure but currently, I feel that `@static` and `@member {Boolean} module:utils/env~env#isGecko` fights with each other. The `#` means that the property should be an instance prop, while the `@static` later tries to convert it into the static prop.

But actually, I don't even know if namespace props should be static or instantiable 😄

AFAIK The doclets with `undocumented: true` are cut out during docs generation.

ma2ciek: But I feel that static props are more accurate.

PR #289, opened 2019-06-24
Add context to `CKEditorError`, change `isCKEditorError()` to `is()`, introduce the `areConnectedThroughProperties()` util

ma2ciek:

Other: Added context as second required argument to the `CKEditorError`'s constructor, changed `isCKEditorError()` method to `is()`. Introduced the `areConnectedThroughProperties()` utility Part of the ckeditor/ckeditor5-watchdog#1.

---

### Additional information

* The added context is required by the Watchdog feature which will be trying to determine which editor crashed in order to restart the proper one. The context might include the editor instance some deep in the JS structure (including objects, array, maps, etc) or the context might be a part accessible from the editor (like `editor.model`, etc.)
* The old `CKEditorError.isCKEditorError()` method was changed to `CKEditorError.is()` and no longer uses the `instanceof` check.

### DIFF
--- a/src/areconnectedthroughproperties.js
+++ b/src/areconnectedthroughproperties.js
@@ -10,18 +10,18 @@
 /* globals EventTarget, Event */
 
 /**
- * Traverses both structures to find out whether there's some object that is shared between both structures.
+ * Traverses both structures to find out whether there is a reference that is shared between both structures.
  *
- * @param {*} str1
- * @param {*} str2
+ * @param {Object|Array} obj1
+ * @param {Object|Array} obj2
  */
-export default function areConnectedThroughProperties( str1, str2 ) {
-	if ( str1 === str2 && isObject( str1 ) ) {
+export default function areConnectedThroughProperties( obj1, obj2 ) {
+	if ( obj1 === obj2 && isObject( obj1 ) ) {
 		return true;
 	}
 
-	const subNodes1 = getSubNodes( str1 );
-	const subNodes2 = getSubNodes( str2 );
+	const subNodes1 = getSubNodes( obj1 );
+	const subNodes2 = getSubNodes( obj2 );
 
 	for ( const node of subNodes1 ) {
 		if ( subNodes2.has( node ) ) {
--- a/src/areconnectedthroughproperties.js
+++ b/src/areconnectedthroughproperties.js
@@ -7,7 +7,7 @@
  * @module utils/arestructuresconnected
  */
 
-/* globals EventTarget */
+/* globals EventTarget, Event */
 
 /**
  * Traverses both structures to find out whether there's some object that is shared between both structures.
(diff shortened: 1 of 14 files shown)


src/ckeditorerror.js
```diff
@@ -32,11 +32,14 @@ export default class CKEditorError extends Error {
 	 * @param {String} message The error message in an `error-name: Error message.` format.
 	 * During the minification process the "Error message" part will be removed to limit the code size
 	 * and a link to this error documentation will be added to the `message`.
+	 * @param {Object|null} context A context of the error by which the {@link module:watchdog/watchdog~Watchdog watchdog}
+	 * is able to determine which editor crashed. It should be an editor instance or a property connected to it. It can be also
+	 * a `null` value if the editor should not be restarted in case of the error (e.g. during the editor initialization).
```
pjasiun: I think you can leave here note that one can use `areConnectedThroughProperties( editor, context )` to check if the object works as the context.

tests/_utils/utils.js
```diff
@@ -6,84 +6,138 @@
...
+export function assertCKEditorError( err, message, editorThatShouldBeFindableFromContext, data ) {
+	if ( typeof message === 'string' ) {
+		message = new RegExp( message );
+	}
+
+	expect( message ).to.be.a( 'regexp', 'Error message should be a string or a regexp.' );
+	expect( err ).to.be.instanceOf( CKEditorError );
+	expect( err.message ).to.match( message, 'Error message does not match the provided one.' );
+
+	// TODO: The `editorThatShouldBeFindableFromContext` is optional but should be required in the future.
```
pjasiun: Follow-up?

src/areconnectedthroughproperties.js
```diff
@@ -0,0 +1,92 @@
...
+ * @module utils/arestructuresconnected
+ */
+
+/* globals EventTarget, Event */
+
+/**
+ * Traverses both structures to find out whether there's some object that is shared between both structures.
+ *
+ * @param {*} str1
+ * @param {*} str2
```
scofalik: Shouldn't it be `{Object}` and maybe we could change names to `obj1` and `obj2`? :)

scofalik: Or maybe `{Object|null}` if that's a case?

ma2ciek: It's hard to be precise in JS types 😅

In TS there are an `object` and an `Object` ( `{}` ) types, where only objects are assignable to the first while all primitive values except `null` and `undefined` can be assigned to the second (all stuff inheriting from the `Object`).

But in our JSDoc system, we use mainly `*` for such types. We use the `Object` only when we accept objects, not primitives. And bear in mind that arrays should be assignable here too 😄

ma2ciek: OTOH I can take your point, that from this function might expect objects and arrays only.

PR #292, opened 2019-07-03
Add feature detection of unicode properties support

msamsel:

Feature: Add feature detection of unicode properties support.

---

### Additional information

* Feature required for ckeditor/ckeditor5-word-count#5
* Feature required for ckeditor/ckeditor5-mention#83


src/featuredetection.js
```diff
@@ -0,0 +1,36 @@
...
+ * @module utils/featuredetection
+ */
+
+/**
+ * Holds feature detection used by the editor.
+ *
+ * @protected
+ * @namespace
+ */
+export default {
```
mlewand: I was wondering whether it should be palced in https://github.com/ckeditor/ckeditor5-utils/blob/cf06347de71b235b564a13016a260af7c191a34d/src/unicode.js - but at the end the linked file is related to unicode handling, while this check is focused on RegExp in context of a given unicode goup handling.

However I'd give it better namespacing, I'm thinking more like: `utils.featuredetection.regexp.isXyzSupported()`.

jodator: @mlewand  to clarify:
```js
import featureDetection from '@ckeditor5-utils/src/utils/featuredetection';

console.log(
    'Regex group is supported:',
    featureDetection.regex.isUnicodePropertySupported()
);
```

ps.: Shouldn't it be `isUnicodePropertyEscapesSupported()`? :laughing:

jodator: @msamsel please create also a PR in mention repo to use this util instaed own feature detection.

src/featuredetection.js
```diff
@@ -0,0 +1,36 @@
...
+	/**
+	 * Indicates whether the current browser supports ES2018 Unicode properties like `\p{P}` or `\p{L}`.
+	 *
```
jodator: Since I'm always searching for this. Please add a link to:

https://www.unicode.org/reports/tr44/#GC_Values_Table

so we get a link to those categories :)

src/featuredetection.js
mlewand: @jodator I was thinking more about:

```js
import regExpFeatureDetection from '@ckeditor5-utils/src/utils/featuredetection/regexp';

console.log(
    'RegExp group is supported:',
    regExpFeatureDetection.isUnicodePropertySupported()
);
```

This would allow us to nicely organize stuff in future.

And it's important for us to use RegExp with **p** consistently, as this is how the type is named in JS.

jodator: >  RegExp with **p** consistently, a

typo, sorry :P

msamsel:
> ps.: Shouldn't it be isUnicodePropertyEscapesSupported()? 😆

@jodator usually in the documentation you can find the name "Unicode property" or just "property". The word "escape" seems to be strictly related to the existence of `\` inside a specific regular expression. I think, in this case, the word "escape" should be skipped, as we don't talk about usage inside regular expression syntax, but rather accessibility of the browser feature.

mlewand:
Also let's try to keep the names as simple as possible 🙂

jodator:
OMG next time I'll use `<joke> ... </joke>` :smile:

msamsel:
> OMG next time I'll use <joke> ... </joke> 😄

Those could be really useful, as it is really hard to notice the sarcasm in the written text, where you expect the feedback ;)

tests/featuredetection/regexp.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+import regExpFeatureDetection from '../../src/featuredetection/regexp';
+
+describe( 'featuredetection', () => {
+	describe( 'isUnicodePropertySupported', () => {
```
Reinmar:
```suggestion
	describe( 'isUnicodePropertySupported()', () => {
```

src/featuredetection/regexp.js
```diff
@@ -0,0 +1,37 @@
...
+	 * @type {Boolean}
+	 */
+	isUnicodePropertySupported: ( function() {
+		let isSupported = false;
+
+		// Feature detection for Unicode properties. Added in ES2018. Currently Firefox and Edge do not support it.
+		// See https://github.com/ckeditor/ckeditor5-mention/issues/44#issuecomment-487002174.
+
+		try {
+			isSupported = 'ć'.search( new RegExp( '[\\p{L}]', 'u' ) ) === 0;
```
Reinmar: Why `new RegExp` and not a literal?

src/featuredetection/regexp.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module utils/featuredetection/regexp
```
Reinmar: Does it have to be entire namespace for regexp? Do we expect to have many of those?

Reinmar: I think that moving this to `env.js` will be cleaner.

src/featuredetection/regexp.js
jodator: @Reinmar I'm investigating this - ESLint is complaining about this.

jodator: I mean that we have to bumpt the ESLint rules used in CKEditor to be able to use `/\p{L}/u` AFAICS.

jodator: OH we've already published the changes and after updating dependencies it works :) So our ESLint config works with this now :tada:

src/featuredetection/regexp.js
mlewand: > Does it have to be entire namespace for regexp? Do we expect to have many of those?

@Reinmar There are several features that are about to come (not widely supported) into JS RegExp:

* s (a.k.a. dotAll) flag
* named groups
* Unicode property escapes (that's in this PR)

Also ES2018 finally added lookahead/lookbehind, luckily both seems to be implemented by all major browser vendors. 🎉

Note that future ES revisions might also expose more RegExp features to further reduce the gap.

src/featuredetection/regexp.js
msamsel: Ok so I tried it, and it's not working with our environment :(
Event when ESLint does not return an error, there is run `acorn` somewhere under the hood which throws a bunch of errors with those regexp literals:
Screenshot 2019-07-18 at 11 21 22

src/featuredetection/regexp.js
msamsel: > There are several features that are about to come (not widely supported) into JS RegExp: (..)

I wonder how many of them are we going to use and will have to detect.

For me having feature detection inside `env.js` seems to be a good solution. The first example is more appealing to me.

Reinmar: @mlewand What would be the problem if we had like 5 regexp-oriented properties in `env`?

We can create new namespaces for every new distinctive thing that may become "plural" with time. But many of them don't become plural and majority doesn't exceed 3 options. If you don't see technical problems for keeping that in `env`, I'd keep that in `env`. They will be easier to find there.

Reinmar: @jodator 

```
       // or just
	features: {
		isRegExpUnicodePropertySupported: isRegExpUnicodePropertySupported()
	}
```

sounds best to me – it's the most flexible option.

src/featuredetection/regexp.js
Reinmar: OK, fair enough. Please leave a comment in the code with an explanation.

jodator:
@msamsel The one thing remains - move the feature detection to the env as

```js
features: {
		isRegExpUnicodePropertySupported: isRegExpUnicodePropertySupported()
	}
```

> I'd keep that in env. They will be easier to find there.

msamsel:
@jodator I've moved feature detection to env.

ma2ciek:
> features: {
		isRegExpUnicodePropertySupported: isRegExpUnicodePropertySupported()
	}

This looks a little bit inconsistent for me, sorry for bumping late into the established discussion.

Shouldn't it be like the following? AFAIK we don't prefix boolean props with `is`, `has`, etc.

```js
features: {
	regExpUnicodePropertySupported: isRegExpUnicodePropertySupported()
}
```

jodator:
> Shouldn't it be like the following? AFAIK we don't prefix boolean props with `is`, `has`, etc.

@ma2ciek dunno, it looks like we do :P

https://github.com/ckeditor/ckeditor5-utils/blob/21f1c4efb028e5f59b0705109ab8d62b536e2b52/src/env.js#L26

https://github.com/ckeditor/ckeditor5-utils/blob/21f1c4efb028e5f59b0705109ab8d62b536e2b52/src/env.js#L34

https://github.com/ckeditor/ckeditor5-utils/blob/21f1c4efb028e5f59b0705109ab8d62b536e2b52/src/env.js#L42

from other files:
https://github.com/ckeditor/ckeditor5-engine/blob/9a40550a7425670264974edc9252f3dcf888ce2f/src/model/element.js#L87

https://github.com/ckeditor/ckeditor5-engine/blob/a58265d1c76c1e2345772861b05770280af13297/src/model/documentselection.js#L680

So we actually prefix boolean values with is/has/are :)

ma2ciek:
>> Shouldn't it be like the following? AFAIK we don't prefix boolean props with is, has, etc.

> @ma2ciek dunno, it looks like we do :P

Wow. I was pretty sure we do the opposite. So let's be consistent.

PS OTOH I don't like this convention because you end up not knowing if the property has a call signature or not.

jodator:
I'm merging this one as other PRs might be closed later on and this will be already used by Mention feature.

PR #293, opened 2019-07-04
t/ckeditor5-ui/515: The `getOptimalPosition()` utility should consider element ancestors with `overflow` to avoid cropping

oleq:

Fix: The `getOptimalPosition()` utility should consider element ancestors with `overflow` to avoid cropping. Closes ckeditor/ckeditor5-ui#515.


Reinmar:
This is just a PoC from @oleq. It requires automated and a lot of manual tests still.

PR #295, opened 2019-07-10
Remove the CKEditor 5 logger and its usage

ma2ciek:

Other: Removed the CKEditor 5 logger and its usage. Part of ckeditor/ckeditor5#383.

---

### Additional information

Requires other connected PRs to be merged first.


src/dom/rect.js
```diff
@@ -82,10 +84,9 @@ export default class Rect {
 				 * @error rect-source-not-in-dom
 				 * @param {String} source The source of the Rect instance.
 				 */
-				log.warn(
+				console.warn( attachLinkToDocumentation(
 					'rect-source-not-in-dom: The source of this rect does not belong to any rendered DOM tree.',
-					{ source }
-				);
+				), { source } );
```
pjasiun: This should be displayed only in the debug mode, this error tells nothing to end users.

src/version.js
```diff
@@ -138,7 +138,9 @@ if ( windowOrGlobal.CKEDITOR_VERSION ) {
 	 *
 	 * @error ckeditor-duplicated-modules
 	 */
-	log.error( 'ckeditor-duplicated-modules: Some CKEditor 5 modules are duplicated.' );
+	console.error( attachLinkToDocumentation(
```
pjasiun: This should be CKEditorError expectation.

src/dom/rect.js
ma2ciek: Maybe the error code/message could be better? Because I think that the warning should be displayed at least.

PR #296, opened 2019-07-11
T/ckeditor5/1151: Allowed passing the content language to `Locale` constructor

oleq:

Feature: Allowed specifying editor content language in `Locale`. Implemented the (UI and content) language direction discovery in `Locale`. Implemented `Locale#uiLanguage`, `Locale#uiLanguageDirection`, `Locale#contentLanguage`, and `Locale#contentLanguageDirection` properties. See ckeditor/ckeditor5#1151.

BREAKING CHANGE: The `Locale.constructor()` arguments syntax has changed. Please refer to the [API documentation](https://ckeditor.com/docs/ckeditor5/latest/api/module_utils_locale-Locale.html#function-constructor) to learn more.

BREAKING CHANGE: The `Locale#language` property has been replaced by `Locale#uiLanguage`. Please refer to the [API documentation](https://ckeditor.com/docs/ckeditor5/latest/api/module_utils_locale-Locale.html#member-uilanguage) to learn more.

---

### Additional information

A piece of https://github.com/ckeditor/ckeditor5/pull/1881.


src/locale.js
```diff
@@ -9,24 +9,66 @@
...
 	/**
 	 * Creates a new instance of the Locale class.
 	 *
-	 * @param {String} [language='en'] The language code in [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) format.
+	 * @param {String} [language='en'] The editor language code in the [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) format.
+	 * @param {String} [contentLanguage] The editor content language code in the
+	 * [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) format.
 	 */
-	constructor( language ) {
+	constructor( language, contentLanguage ) {
```
Reinmar: This starts to look bad. What does `new Locale( 'en', 'pl' )` do? Also, we even talked about allowing configuring the content direction separately, so that'd be a 3rd param.

I think it's a good moment to turn this into an object. I'd still support the old `new Locale( lang )` form but log a deprecation warning, though. We'll be able to clean this up shortly.

Reinmar: Oooor... we can say that if you do `Locale( 'en' )` then you set both things to `en`. If you want to set them independently, pass an object: `Locale( { ui: 'en', content: 'ar' } )`. 

This would also affect the config of the editor:

```js
.create( foo, {
    language: 'en' // you set both
} );
```

vs

```js
.create( foo, {
    language: {
      ui: 'en',
      content: 'ar'
    }
} );
```

There would be 2 problems, though:

* the current `locale.language` property :|,
* `config.language.contentDirection` would look bad too

So, perhaps going with independent properties for them is better. Still, we could support `Locale( 'en' )` as a shorthand for: Locale( { language: 'en', contentLanguage: 'en', contentDirection: 'ltr' } )`. So, in other words – perhaps it does not make sense to deprecate the current short form because it's quite useful.

oleq: >  we can say that if you do Locale( 'en' ) then you set both things to en

This is how it works now in this PR. `Locale( 'en' )` will set `language` and `contentLanguage` to `en`.

> This would also affect the config of the editor:

While I'm OK with deprecating `Locale( '...' )` and using `Locale( { language: '...', [ contentLanguage: '...' ] )` instead, I'm not sure we should touch the editor config. 

----

Or maybe let me put it this way: we should go all-in (one big breaking change) or not go at all. If we go all–in we're gonna have:

```
.create( foo, {
    language: 'en' // backward compat that sets both
} );
```

AND 

```
.create( foo, {
    language: {
      ui: 'en',
      content: 'ar'
    }
} );
```

AND

`Locale( { uiLanguage: 'en', contentLanguage: 'ar' } )` 

AND 

`Locale#uiLanguage` + `Locale#contentLanguage`.

oleq: Random thought: what if someone created a multi–root editor that has a different content language for each root? Use case: two editables side–by–side where the user translates the content (writes the same thing in 2 or more languages at the time).

Reinmar:
I already forgot that, but didn't we want to leave `Locale#language` getter with a warning to not break the apps?

PR #307, opened 2019-10-07
Allow creating namespaced domelements

### LINKED ISSUE #306: Support creation of namespaced elements
This relates to https://github.com/ckeditor/ckeditor5/issues/1842.

Currently it's not possible to use createElement to create working namespaced elements like SVGs. 

PR is on its way.

Sebobo:
Fix: Create valid namespaced elements. Closes #306.

---

### Additional information

* Relates to https://github.com/ckeditor/ckeditor5/issues/1842

### DIFF
--- a/src/dom/createelement.js
+++ b/src/dom/createelement.js
@@ -26,7 +26,8 @@ import { isString } from 'lodash-es';
  * @returns {Element} Created element.
  */
 export default function createElement( doc, name, attributes = {}, children = [] ) {
-	const element = doc.createElement( name );
+	const namespace = attributes instanceof Object ? attributes.xmlns : null;
+	const element = namespace ? doc.createElementNS( namespace, name ) : doc.createElement( name );
 
 	for ( const key in attributes ) {
 		element.setAttribute( key, attributes[ key ] );
--- a/tests/dom/createelement.js
+++ b/tests/dom/createelement.js
@@ -23,6 +23,15 @@ describe( 'createElement', () => {
 		expect( p.getAttribute( 'class' ) ).to.equal( 'foo' );
 	} );
 
+	it( 'should create element with namespace', () => {
+		const namespace = 'http://www.w3.org/2000/svg';
+		const svg = createElement( document, 'svg', { xmlns: namespace } );
+
+		expect( svg.tagName.toLowerCase() ).to.equal( 'svg' );
+		expect( svg.getAttribute( 'xmlns' ) ).to.equal( namespace );
+		expect( svg.createSVGRect instanceof Function ).to.equal( true );
+	} );
+
 	it( 'should create element with child text node', () => {
 		const p = createElement( document, 'p', null, 'foo' );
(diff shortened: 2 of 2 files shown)


mlewand:
See ckeditor5#2088.

src/dom/createelement.js
```diff
@@ -26,7 +26,8 @@ import { isString } from 'lodash-es';
  * @returns {Element} Created element.
  */
 export default function createElement( doc, name, attributes = {}, children = [] ) {
-	const element = doc.createElement( name );
+	const namespace = attributes instanceof Object ? attributes.xmlns : null;
+	const element = namespace ? doc.createElementNS( namespace, name ) : doc.createElement( name );
```
Reinmar: I think that you don't need to check whether `attributes` is an object. You can simply assume it is (which happens later in this function anyway).

tests/dom/createelement.js
```diff
@@ -23,6 +23,15 @@ describe( 'createElement', () => {
 		expect( p.getAttribute( 'class' ) ).to.equal( 'foo' );
 	} );
 
+	it( 'should create element with namespace', () => {
+		const namespace = 'http://www.w3.org/2000/svg';
+		const svg = createElement( document, 'svg', { xmlns: namespace } );
+
+		expect( svg.tagName.toLowerCase() ).to.equal( 'svg' );
+		expect( svg.getAttribute( 'xmlns' ) ).to.equal( namespace );
+		expect( svg.createSVGRect instanceof Function ).to.equal( true );
```
Reinmar: I think this should work:

```suggestion
		expect( svg.createSVGRect ).to.be.a( 'function' );
```

Reinmar:
Thanks for the PR. I found two small issues and I think that's it.

src/dom/createelement.js
Reinmar: > You can simply assume it is (which happens later in this function anyway).

Actually, I thought there's `for-of` loop next, which would verify that `attributes` is an object. But it's `for-in` so it'll work fine with `undefined` and `null`. So, if other tests or the reality requires that you handle `attributes` being `null` or `undefined, then this should be fine:

```js
const namespace = attributes && attributes.xmlns;
```

Sebobo: I actually had the problem that in some tests attributes was null and my code failed.

tests/dom/createelement.js
Sebobo: Cool, thanks for the hint!

src/dom/createelement.js
Sebobo: So based on that, should I still change something or leave it like that?

Reinmar: I think you could simplify this to:

```js
const namespace = attributes && attributes.xmlns;
```

I'll try this on `master` after merging your PR.

Reinmar:
Thanks again! ❤

PR #312, opened 2019-10-28
Use other style in normalizeHtml tests.

jodator:

Tests: Align tests to the changes in style normalization.

---

### Additional information

Sub-pr for: https://github.com/ckeditor/ckeditor5-engine/pull/1807.

* I've used safer `padding` style instead of the `border` which is output as `border-top, -right,`etc.

### DIFF
--- a/tests/_utils-tests/normalizehtml.js
+++ b/tests/_utils-tests/normalizehtml.js
@@ -8,29 +8,29 @@ import normalizeHtml from '../../tests/_utils/normalizehtml';
 describe( 'utils', () => {
 	describe( 'normalizeHtml', () => {
 		it( 'should sort attributes', () => {
-			const actual = '<a style="border:1px" class="" href="file://"></a>';
-			const expected = '<a class="" href="file://" style="border:1px"></a>';
+			const actual = '<a style="padding:1px" class="" href="file://"></a>';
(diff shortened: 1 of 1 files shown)


Reinmar:
This PR doesn't seem to be needed anymore as we apparently output a shorthand option. 

However, this made me realise the most unfortunate effect of having StylesMap's options  in form of a singleton. We do affect other test files when loading a processor in one of them... And I can't think on a solution for that ;/

Reinmar:
> This PR doesn't seem to be needed anymore as we apparently output a shorthand option.

OK, that's untrue. It's even worse – it may or may not be needed depending on whether table or engine tests are run before it. If engine, it's ok cause apparently engine tests do stub border normalization to something that gives 1:1 results. But table tests use a real style processor and it breaks this test ;O.

Reinmar:
> This PR doesn't seem to be needed anymore as we apparently output a shorthand option.

OK, that's untrue. It's even worse – it may or may not be needed depending on whether table or engine tests are run before it. If engine, it's ok cause apparently engine tests do stub border normalization to something that gives 1:1 results. But table tests use a real style processor and it breaks this test ;O.

Reinmar:
I created a followup for the border processor: [https://github.com/ckeditor/ckeditor5/issues/6090](https://github.com/ckeditor/ckeditor5/issues/6090).

PR #316, opened 2019-12-03
Added iterator interface to the Config class.

oskarwrobel:

Feature: Added iterator interface to the `Config` class. Part of ckeditor/ckeditor5#5891.

---

### Additional information

Required by the https://github.com/ckeditor/ckeditor5-core/pull/204. `Editor#config` needs to be extended by the `Context#config`.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -114,14 +114,14 @@ export default class Config {
 	}
 
 	/**
-	 * Iterable interface.
-	 *
 	 * Iterates over all top level configuration names.
 	 *
 	 * @returns {Iterable.<String>}
 	 */
-	[ Symbol.iterator ]() {
-		return Object.keys( this._config )[ Symbol.iterator ]();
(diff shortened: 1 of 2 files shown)


src/config.js
```diff
@@ -113,6 +113,17 @@ export default class Config {
...
 	}
 
+	/**
+	 * Iterable interface.
+	 *
+	 * Iterates over all top level configuration names.
+	 *
+	 * @returns {Iterable.<String>}
+	 */
+	[ Symbol.iterator ]() {
```
Reinmar: If it's iterating over names I'd create it under `Config#names()` or `Config#keys()` depending on how we call this thing. Similar to `Map#keys()` vs `Map#iterator` where the former iterates over prop names and the other over all entries.

oskarwrobel: Done. I added a `Config#names()` method.

PR #321, opened 2020-01-29
WIP: Implemented new gettext-like functions.

phaux:

Feature: Implemented new gettext-like functions.

Closes: ckeditor/ckeditor5#6158

### DIFF
--- a/src/locale.js
+++ b/src/locale.js
@@ -76,25 +76,69 @@ export default class Locale {
 		this.contentLanguageDirection = getLanguageDirection( this.contentLanguage );
 
 		/**
-		 * Translates the given string to the {@link #uiLanguage}. This method is also available in
+		 * Translates the given string to the {@link #uiLanguage}.
+		 *
+		 * This method is also available in
 		 * {@link module:core/editor/editor~Editor#t} and {@link module:ui/view~View#t}.
 		 *
-		 * The strings may contain placeholders (`%<index>`) for values which are passed as the second argument.
+		 * The string may contain placeholders (`%<index>`) for given dynamic values.
 		 * `<index>` is the index in the `values` array.
 		 *
-		 *		editor.t( 'Created file "%0" in %1ms.', [ fileName, timeTaken ] );
-		 *
-		 * This method's context is statically bound to Locale instance,
-		 * so it can be called as a function:
-		 *
-		 *		const t = this.t;
-		 *		t( 'Label' );
+		 *		editor.t( 'Created file "%0" in %1ms', [ fileName, time ] );
 		 *
 		 * @method #t
-		 * @param {String} str The string to translate.
+		 * @param {String} msgid The string to translate.
 		 * @param {String[]} [values] Values that should be used to interpolate the string.
 		 */
-		this.t = ( ...args ) => this._t( ...args );
+		this.t = ( msgid, values ) => this._t( msgid, values );
+
+		/**
+		 * Translates the given string to the {@link #uiLanguage}.
+		 *
+		 * This method allows passing the context as the first argument.
+		 * A context is used to provide a description or example of how the translated string will be used.
+		 * An unique context makes the whole translation unique and should be presented to the translator.
+		 *
+		 * The string may contain placeholders (`%<index>`) for given dynamic values.
+		 * `<index>` is the index in the `values` array.
+		 *
+		 *		editor.ct( 'Created file "x"', 'file "%0"', [ fileName ] );
+		 *
+		 * If at least one of the placeholder values is not a string or number,
+		 * then an array of all the literal and dynamic parts will be returned
+		 *
+		 * @method #ct
+		 * @template T
+		 * @param {string} msgctxt The context string.
+		 * @param {string} msgid The string to translate.
+		 * @param {T[]} [values] Values that should be used to interpolate the string.
+		 * @returns {T extends object ? Array<T | string> : string}
+		 */
+		this.ct = ( msgctxt, msgid, values ) => this._ct( msgctxt, msgid, values );
+
+		/**
+		 * Translates the given plural string to the {@link #uiLanguage}.
+		 *
+		 * This method is also available in
+		 * {@link module:core/editor/editor~Editor#t} and {@link module:ui/view~View#t}.
+		 *
+		 * This method allows passing the context as the first argument.
+		 * A context is used to provide a description or example of how the translated string will be used.
+		 * An unique context makes the whole translation unique and should be presented to the translator.
+		 *
+		 * The strings may contain a placeholder (`%0`) for a given dynamic value.
+		 * The value is also used for selecting the correct plural form.
+		 *
+		 *		editor.ctn( 'Created n files', '%0 file' '%0 files' fileCount );
+		 *
+		 * @method #ctn
+		 * @param {string} msgctxt The context string.
+		 * @param {string} msgid Singular form of a string to translate.
+		 * @param {string} msgidPlural Plural form of a string to translate.
+		 * @param {number} value Value that should be used to interpolate the string.
+		 * @returns {string}
+		 */
+		this.ctn = ( msgctxt, msgid, msgidPlural, value ) => this._ctn( msgctxt, msgid, msgidPlural, value );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


src/locale.js
```diff
@@ -123,19 +167,45 @@ export default class Locale {
...
+		return parts.join( '' );
+	}
 
-		return translatedString;
+	/**
+	 * Base for the {@link #ctn} method.
+	 * @private
+	 */
+	_ctn( msgctxt = '', msgid = '', msgidPlural = '', value = 0 ) {
+		return this._ct( msgctxt, value == 1 ? msgid : msgidPlural, [ value ] );
```
scofalik: Will it work? What kind of JSON structure do you expect if `_ct` is used like that? How `translate` works? Assume that the language is Polish (translation keys are in English, of course).

Also, we are giving up a possibility to use multiple placeholders, which is not ideal. How about `_ctn( msgctxt = '', msgid = '', msgidPlural = '', quantity = 0, values = null )` and if `values` is not set, then it becomes `[ quantity ]?`

Example usage: `ctn( 'Sent x of y files', 'file', '%0 of %1 files', 4, [ 1, 4 ] );`

src/locale.js
```diff
@@ -123,19 +167,45 @@ export default class Locale {
...
+			}
+			parts.push( msg.substring( prevIndex, match.index ) );
+			parts.push( values[ match[ 1 ] ] );
+		}
 
-		if ( values ) {
-			translatedString = translatedString.replace( /%(\d+)/g, ( match, index ) => {
-				return ( index < values.length ) ? values[ index ] : match;
-			} );
+		if ( values.some( value => typeof value == 'object' ) ) {
```
scofalik: I could argue that Boolean type can be casted to a string, however maybe we could also check if a value is a function? Maybe there would be uses for this functionallity.

src/locale.js
```diff
@@ -76,25 +76,69 @@ export default class Locale {
...
+		 * A context is used to provide a description or example of how the translated string will be used.
+		 * An unique context makes the whole translation unique and should be presented to the translator.
+		 *
+		 * The string may contain placeholders (`%<index>`) for given dynamic values.
+		 * `<index>` is the index in the `values` array.
+		 *
+		 *		editor.ct( 'Created file "x"', 'file "%0"', [ fileName ] );
+		 *
+		 * If at least one of the placeholder values is not a string or number,
+		 * then an array of all the literal and dynamic parts will be returned
```
scofalik: . missing at the end of the sentence.

Please expand with an example. It was difficult for me to understand what this means and how I could use it.

src/locale.js
```diff
@@ -76,25 +76,69 @@ export default class Locale {
...
+		 * {@link module:core/editor/editor~Editor#t} and {@link module:ui/view~View#t}.
+		 *
+		 * This method allows passing the context as the first argument.
+		 * A context is used to provide a description or example of how the translated string will be used.
+		 * An unique context makes the whole translation unique and should be presented to the translator.
+		 *
+		 * The strings may contain a placeholder (`%0`) for a given dynamic value.
+		 * The value is also used for selecting the correct plural form.
+		 *
+		 *		editor.ctn( 'Created n files', '%0 file' '%0 files' fileCount );
```
scofalik: When you fix docs for `ct` please check this function docs against `ct` docs. Some stuff is missing here, AFAICT.

src/locale.js
```diff
@@ -76,25 +76,69 @@ export default class Locale {
...
-		 * so it can be called as a function:
-		 *
-		 *		const t = this.t;
-		 *		t( 'Label' );
+		 *		editor.t( 'Created file "%0" in %1ms', [ fileName, time ] );
 		 *
 		 * @method #t
-		 * @param {String} str The string to translate.
+		 * @param {String} msgid The string to translate.
 		 * @param {String[]} [values] Values that should be used to interpolate the string.
```
scofalik: Array.<String>

src/locale.js
scofalik: I think that translate function could take quantity as an additional parameter and return a proper string from the JSON language file.

src/locale.js
```diff
@@ -123,19 +167,45 @@ export default class Locale {
 
 	/**
 	 * Base for the {@link #t} method.
-	 *
 	 * @private
 	 */
-	_t( str, values ) {
-		let translatedString = translate( this.uiLanguage, str );
+	_t( msgid = '', values = [] ) {
+		return translate( this.uiLanguage, msgid )
```
scofalik: Why not have the same functionallity here and in `_ct` and `_ctn` (splitting to array?).

We could move lines 183-200 to a seprate function and use them in `_t`, `_ct` and `_ctn`.

scofalik:
Apart from what I commented:

1.  Add additional blank line before `if`s and after `if`s.
2.  Docs:
    1.  use capital letters for types (`string` => `String`)
    2.  use `Array.<T>` instead of `T[]`
3.  Tests are missing.

src/locale.js
phaux: The plural function usually takes only one parameter. I will implement proper plural selection after I change the webpack plugin

src/locale.js
```diff
@@ -76,25 +76,48 @@ export default class Locale {
...
-		 * @param {String[]} [values] Values that should be used to interpolate the string.
+		 * @method #ctn
+		 * @param {String} msgctxt The context string.
+		 * @param {String} msgid Singular form of a string to translate.
+		 * @param {String} msgidPlural Plural form of a string to translate.
+		 * @param {Number} quantity
+		 * @returns {Message}
 		 */
-		this.t = ( ...args ) => this._t( ...args );
+		this.ctn = ( msgctxt, msgid, msgidPlural, quantity ) => this._ctn( msgctxt, msgid, msgidPlural, quantity );
```
scofalik: How about we name this function `ctq`? Context-translation-quantity?

src/locale.js
```diff
@@ -76,25 +76,48 @@ export default class Locale {
...
 		/**
-		 * Translates the given string to the {@link #uiLanguage}. This method is also available in
-		 * {@link module:core/editor/editor~Editor#t} and {@link module:ui/view~View#t}.
+		 * Translates the given string to the {@link #uiLanguage}.
 		 *
-		 * The strings may contain placeholders (`%<index>`) for values which are passed as the second argument.
-		 * `<index>` is the index in the `values` array.
+		 * @method #t
+		 * @param {String} msgid The string to translate.
+		 * @returns {Message}
```
scofalik: Does it return `Message` or `String`?

src/locale.js
```diff
@@ -146,3 +184,85 @@ export default class Locale {
...
+	 */
+	toString() {
+		return this.message;
+	}
+
+	/**
+	 * Get the interpolated message as string.
+	 * This method will replace the placeholders in the message with provided values.
+	 *
+	 * @example
```
scofalik: I haven't seen us using those tags before, so make sure they are correctly parsed by Umberto into API docs.

src/locale.js
```diff
@@ -146,3 +184,85 @@ export default class Locale {
...
+		return this.message.replace( /%(\d+)/g, ( match, index ) => values[ parseInt( index, 10 ) ] );
+	}
+
+	/**
+	 * Get the interpolated message as an array of static and dynamic parts.
+	 * This method splits the message at every placeholder
+	 * and returns an array of the original string parts with provided values interspersed.
+	 * The values can be of any type and won't be stringified.
+	 *
+	 * @example
```
scofalik: I'd like to see an example with non-strings here.

src/locale.js
scofalik: I think it can return `Message`, if it is stringify-able easily. But then we need to change `t()` function definition.

src/locale.js
```diff
@@ -146,3 +184,85 @@ export default class Locale {
 function getLanguageDirection( languageCode ) {
 	return RTL_LANGUAGE_CODES.includes( languageCode ) ? 'rtl' : 'ltr';
 }
+
+/**
+ * A translated message returned from one of the {@link Locale}'s methods.
+ */
+export class Message {
+	/**
+	 * @hidden
```
ma2ciek: What's the `hidden` tag? I'm afraid we don't use it anywhere. And I couldn't find it in [https://jsdoc.app/](https://jsdoc.app/)  too.

src/locale.js
```diff
@@ -146,3 +184,85 @@ export default class Locale {
...
+	 * The values can be of any type and won't be stringified.
+	 *
+	 * @example
+	 *
+	 * 		const msg = ct( 'Replace text', 'Replace %0 with %1' )
+	 * 			.formatToParts( 'foo', 'bar' )
+	 *
+	 * 		expect( msg ).to.equal( [ 'Replace ', 'foo', ' with ', 'bar' ] )
+	 *
+	 * @template T
```
ma2ciek: I'm also unsure about the generic `@template` Currently we don't use it anywhere. I know it's a very valuable thing, though it can be unsupported currently by Uberto and our JSDoc plugins 😞

ma2ciek:
Closing in favor of #334.

PR #322, opened 2020-02-04
i/6188: The `uid()` helper should be a lot faster

oleq:

Internal: The `uid()` helper should be a lot faster. Closes ckeditor/ckeditor5#6188.

### DIFF
--- a/src/uid.js
+++ b/src/uid.js
@@ -7,18 +7,53 @@
  * @module utils/uid
  */
 
+// A hash table of hex numbers to avoid using toString() in uid() which is costly.
+// [ '00', '01', '02', ..., 'fe', 'ff' ]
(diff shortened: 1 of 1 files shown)


Reinmar:
Since we were working on it in the context of https://github.com/ckeditor/ckeditor5/issues/6161#issuecomment-581384534 I'd like to know its effect on that scenario. Could you compare before/after performance of loading that UI?

oleq:
Nothing spectacular. Around 35ms on total 350ms when switching categories from Arrows->All (~10%).

Then I checked the http://localhost:8125/ckeditor5/tests/manual/performance/richeditor.html test and I created the editor on button click. The difference is 965ms vs 895ms (around 7%).

It's not like the editor is blazingly fast now but this is something.

PR #324, opened 2020-02-13
Introduced copying default configuration in the `Config` class to make the watchdog feature work correctly

ma2ciek:

Internal: Introduced copying default configuration in the `Config` class to make the watchdog feature work correctly. Part of ckeditor/ckeditor5#6219.

---

### Additional information

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -30,7 +30,9 @@ export default class Config {
 
 		// Set default configuration.
 		if ( defaultConfigurations ) {
-			this.define( defaultConfigurations );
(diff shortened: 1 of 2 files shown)


ma2ciek:
I've added a test for sharing props between instances. I've to do it via utility as the `config._config` is a private field and `config.get()` clones the returned value, so there's no other way to check it.

Reinmar:
> I've added a test for sharing props between instances. I've to do it via utility as the `config._config` is a private field and `config.get()` clones the returned value, so there's no other way to check it.

OK, makes sense.

PR #327, opened 2020-03-24
Fixed fastDiff for multi-byte unicode sequences.

niegowski:

Fix: Fixed fastDiff for multi-byte unicode sequences. Closes ckeditor/ckeditor5#3147. Closes ckeditor/ckeditor5#6495.

---

### Additional information
* ckeditor/ckeditor5-typing#228

### DIFF
--- a/src/fastdiff.js
+++ b/src/fastdiff.js
@@ -100,8 +100,12 @@ export default function fastDiff( a, b, cmp, atomicChanges = false ) {
 		return a === b;
 	};
 
-	// Transform text or any iterable into arrays for easier, consistent processing.
-	// Array.from was used here but it generated incorrect results for multi-byte unicode sequences.
+	// Convert the string (or any array-like object - eg. NodeList) to an array by using the slice() method because,
+	// unlike Array.from(), it returns array of UTF-16 code units instead of the code points of a string.
+	// One code point might be a surrogate pair of two code units. All text offsets are expected to be in code units.
+	// See ckeditor/ckeditor5#3147.
+	//
+	// We need to make sure here that fastDiff() works identical to diff().
 	if ( !Array.isArray( a ) ) {
 		a = Array.prototype.slice.call( a );
 	}
--- a/src/fastdiff.js
+++ b/src/fastdiff.js
@@ -101,12 +101,13 @@ export default function fastDiff( a, b, cmp, atomicChanges = false ) {
 	};
 
 	// Transform text or any iterable into arrays for easier, consistent processing.
+	// Array.from was used here but it generated incorrect results for multi-byte unicode sequences.
 	if ( !Array.isArray( a ) ) {
-		a = Array.from( a );
+		a = Array.prototype.map.call( a, c => c );
 	}
 
 	if ( !Array.isArray( b ) ) {
-		b = Array.from( b );
+		b = Array.prototype.map.call( b, c => c );
 	}
 
 	// Find first and last change.
--- a/src/fastdiff.js
+++ b/src/fastdiff.js
@@ -103,11 +103,11 @@ export default function fastDiff( a, b, cmp, atomicChanges = false ) {
 	// Transform text or any iterable into arrays for easier, consistent processing.
 	// Array.from was used here but it generated incorrect results for multi-byte unicode sequences.
 	if ( !Array.isArray( a ) ) {
-		a = Array.prototype.map.call( a, c => c );
+		a = Array.prototype.slice.call( a );
 	}
 
 	if ( !Array.isArray( b ) ) {
-		b = Array.prototype.map.call( b, c => c );
+		b = Array.prototype.slice.call( b );
 	}
 
 	// Find first and last change.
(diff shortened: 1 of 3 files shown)


Reinmar:
Looking at the tests, a single emoji is treated as 2 characters? I'd rather prefer if it was considered as a single item. I think that this is more in line with how we think about them. When you replace :family\_man\_girl:  with :family\_man\_girl\_boy: you want to have one remove and one insert as the whole thing was replaced

WDYT, @scofalik?

Besides that, we need an integration test in ckeditor5-typing that the original issue (typing) does not occur any more.

Reinmar:
Also, I'd love to have a similar set of tests for both `fastDiff()` and `diff()` to verify that both treat those characters in identical ways. Those test should be more advanced (consider more cases),

Reinmar:
Uh, I've just realise that the way how the diff functions work (index-based) does not allow to treat a multi-byte characters as a single item. That's a bummer ;/

src/fastdiff.js
```diff
@@ -101,12 +101,13 @@ export default function fastDiff( a, b, cmp, atomicChanges = false ) {
 	};
 
 	// Transform text or any iterable into arrays for easier, consistent processing.
+	// Array.from was used here but it generated incorrect results for multi-byte unicode sequences.
 	if ( !Array.isArray( a ) ) {
-		a = Array.from( a );
+		a = Array.prototype.map.call( a, c => c );
```
pomek: Can't be `Array.from( a, c => c );` [according to docs](https://developer.mozilla.org/pl/docs/Web/JavaScript/Referencje/Obiekty/Array/from)?

scofalik: No, `Array#from` works differently for multi-byte characters.

niegowski:
Additional tests: ckeditor/ckeditor5-typing#228

Reinmar:
cc @ckeditor/qa-team  Could you test this PR? I'm interested in emoji support, IME on various platforms and perhaps something like spellchecking too. Please make sure to test how the thing becomes with long and short paragraphs.

tests/diff.js
```diff
@@ -67,4 +67,88 @@ describe( 'diff', () => {
...
+	describe( 'with multi-byte unicode', () => {
+		describe( 'simple emoji - single unicode code point', () => {
+			// 🙂 = '\ud83d\ude42' = 2 chars
+			const emojiLength = '🙂'.length;
+			const emojiDiffInsert = new Array( emojiLength ).fill( 'insert' );
+			const emojiDiffEqual = new Array( emojiLength ).fill( 'equal' );
+			const emojiDiffDelete = new Array( emojiLength ).fill( 'delete' );
+
+			it( 'should properly handle emoji insertion', () => {
+				expect( diff( 'abc', 'ab🙂c' ) ).to.deep.equals( [ 'equal', 'equal', ...emojiDiffInsert, 'equal' ] );
```
Reinmar: `to.deep.equal` to make it grammatical

src/fastdiff.js
```diff
@@ -101,12 +101,13 @@ export default function fastDiff( a, b, cmp, atomicChanges = false ) {
 	};
 
 	// Transform text or any iterable into arrays for easier, consistent processing.
+	// Array.from was used here but it generated incorrect results for multi-byte unicode sequences.
```
Reinmar: I'd rather explain why we use the cloning mechanism of of arrays over using Array.from. Something like:

> We convert the string to an array by using the `slice()` method because, unlike Array.from(), it returns single-byte array items. See ckeditor/ckeditor5#3147.
> 
> We need to make sure here that `fastDiff()` works identical to `diff()` here.

tests/diff.js
```diff
@@ -67,4 +67,88 @@ describe( 'diff', () => {
...
+
+			it( 'should properly handle insertion to string containing emoji', () => {
+				expect( diff( 'ab🙂cd', 'ab🙂cde' ) ).to.deep.equals( [ 'equal', 'equal', ...emojiDiffEqual, 'equal', 'equal', 'insert' ] );
+			} );
+
+			it( 'should properly remove emoji', () => {
+				expect( diff( 'a🙂b', 'ab' ) ).to.deep.equals( [ 'equal', ...emojiDiffDelete, 'equal' ] );
+			} );
+
+			it( 'should properly replace emoji', () => {
```
Reinmar: I'm missing a case with replacing one emoji with another one.

tests/diff.js
Reinmar: Frankly speaking, this `emojiDiffInsert` makes these tests much harder to read than they need to be. IMO, assertions should be as "raw" as possible so you don't need to jump to e.g. a definition of a variable to understand them. Here, I needed to scroll up to those consts to understand one of the tests (I started reading them from the bottom). Plus, I needed to check whether I'm right that these consts contain 2 items.

Reinmar: OTOH, I can see that in the ZWJ tests we have 5 items in such consts, so it's more justified here. So I take my above comment back. It's fine the way you coded it.

tests/diff.js
Reinmar: Another test I miss is for several emojis one after another. It creates some additional cases. I don't expect those to ever break, unless we'll start doing some weird code optimizations in diff/fastDiff, so those tests may not be essential, but it's easy to have them just in case.

Reinmar:
The code looks good. I like the tests for their completeness. There's indeed a number of cases to cover and I think there's just one or two I missed.

Mgsy:
We've tested multiple cases and everything seems to be fine 👌

PR #332, opened 2020-04-10
i/6573: The `getPositionedAncestor()` helper should use `offsetParent` instead of `getComputedStyle()`

oleq:

Other: The `getPositionedAncestor()` helper should use `offsetParent` instead of `getComputedStyle()` for performance reasons. Closes ckeditor/ckeditor5#6573.

MINOR BREAKING CHANGE: The `getPositionedAncestor()` helper will no longer return the passed element when it is positioned. 

---

### Additional information

**When closed, it must be followed by an update and a [merge](https://github.com/cksource/ckeditor5-comments/pull/187) in the comments plugin.**

**An important notice: The `getPositionedAncestor()` helper is used only by the `getOptimalPosition()` helper at this moment.**

OK, so here's what happened in this PR:

1. I used `offsetParent` instead of `getComputedStyle()`, which results in a massive performance gain. This is how old vs new implementation performs when executed 100 times on a DOM tree of 100 ancestors:

   Screenshot 2020-04-10 at 09 30 02

   I think the responsiveness of the UI should benefit from it because `getOptimalPosition()` is called dozens of times, for instance, when you resize the viewport, to re-position UI balloons.

2. In the old implementation the helper also considered the passed element (e.g. if it had `position: absolute`). It was wrong on many levels:
   * It wasn't documented.
   * It contradicted the name of the helper.
   * We didn't benefit from that fact.

   Dropping this behaviour allowed me to get rid of `getComputedStyle()`. And things make sense now.

3. All changes made to `position.js` (`getOptimalPosition()`) and its tests are purely to move from mocked to real DOM elements. Using mocks no longer works when `getPositionedAncestor()` uses `offsetParent`.

### DIFF
--- a/src/dom/getpositionedancestor.js
+++ b/src/dom/getpositionedancestor.js
@@ -16,29 +16,13 @@ import global from './global';
  * @returns {HTMLElement|null}
  */
 export default function getPositionedAncestor( element ) {
-	while ( element && element.tagName.toLowerCase() != 'html' ) {
-		if ( global.window.getComputedStyle( element ).position != 'static' ) {
-			return element;
-		}
-
-		element = element.parentElement;
+	if ( !element || !element.parentNode ) {
+		return null;
 	}
 
-	return null;
+	if ( element.offsetParent === global.document.body ) {
+		return null;
+	}
+
+	return element.offsetParent;
 }
-
-// export default function getPositionedAncestor( element ) {
-// 	if ( !element || !element.parentNode ) {
-// 		return null;
-// 	}
-
-// 	if ( element.style.position && element.style.position !== 'static' ) {
-// 		return element;
-// 	}
-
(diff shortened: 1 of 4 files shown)


Reinmar:
Good catch. I'd like to see, though, how it improves the performance in some real scenarios. Could you record timelines for cases where this code is used?

oleq:
As you probably expected the real-life gains are not so spectacular 😛

In Chrome it is around ~5% performance gain for the method (which execution is negligible in the bigger picture, BTW) and ~2fps more while re-positioning balloons a lot (e.g. scrolling/resizing the window) ([a nice answer on SO BTW](https://stackoverflow.com/questions/48079661/how-to-get-the-average-fps-in-chrome-devtools)).

In Firefox I couldn't spot any difference, any gain or loss was lost in the noise.

PR #334, opened 2020-04-17
Provide support for translating plural forms

ma2ciek:

Feature: Provided support for plural forms internalization. Part of ckeditor/ckeditor5#6526.

MINOR BREAKING CHANGE: The `translate` function from the `translation-service` was marked as protected. See #334.
MINOR BREAKING CHANGE: The format of translations added to the editor has been changed. If you use `window.CKEDITOR_TRANSLATIONS` please see #334.

---

### Additional information

The `translate` function in the `translation-service` was marked as protected and renamed to `_translate`. All features depending on the translation mechanism should use the `Locale#t()` function instead.

The new format of the  `window.CKEDITOR_TRANSLATIONS`:
```js
{
   pl: {
      dictionary: {
         'Cancel': 'Anuluj',
         'Add space': [ 'Dodaj spację', 'Dodaj %0 spacje', 'Dodaj %0 spacji' ]
      }, 
      // A function that returns the plural form index.
      // Actually it's more like n => n == 1 ? 0 : n % 10 >= 2 && n % 10 <= 4 && ( n % 100 < 10 || n % 100 >= 20 ) ? 1 : 2
      getPluralForm: n => n !==1
   }
   // other languages...
}
```

### DIFF
--- a/src/locale.js
+++ b/src/locale.js
@@ -79,7 +79,7 @@ export default class Locale {
 		 * Translates the given message to the {@link #uiLanguage}. This method is also available in
 		 * {@link module:core/editor/editor~Editor#t} and {@link module:ui/view~View#t}.
 		 *
-		 * This method's context is statically bound to the `Locale` instance and should be called as a function:
+		 * This method's context is statically bound to the `Locale` instance and **always should be called as a function**:
 		 *
 		 *		const t = locale.t;
 		 *		t( 'Label' );
@@ -92,7 +92,8 @@ export default class Locale {
 		 *		t( 'Created file "%0" in %1ms.', [ fileName, timeTaken ] );
 		 *
 		 * A Message can provide a plural form using the `plural` property and a value - that should be always the first element
-		 * of the `values` array based on which the plural form of the target language should be picked.
+		 * of the `values` array based on which the plural form of the target language should be picked. That property value will
+		 * be used as a default plural translation when the translation for the target language will be missing.
 		 *
 		 *		t( { string: 'Add a space', plural: 'Add %0 spaces' }, [ spaces ] );
 		 *		t( { string: '%1 a space', plural: '%1 %0 spaces' }, [ spaces, 'Add' ] );
@@ -107,6 +108,7 @@ export default class Locale {
 		 * @method #t
 		 * @param {String|module:utils/translation-service~Message} message A message that will be localized.
 		 * @param {Array.<String>} [values] Values that should be used to interpolate the string.
+		 * @returns {String}
 		 */
 		this.t = ( message, values ) => this._t( message, values );
 	}
@@ -138,8 +140,9 @@ export default class Locale {
 	/**
 	 * Base for the {@link #t} method.
 	 *
-	 * @param {module:utils/translation-service~Message|String} message
+	 * @param {String|module:utils/translation-service~Message} message
 	 * @param {Array.<String>} [values]
+	 * @returns {String}
 	 * @private
 	 */
 	_t( message, values = [] ) {
--- a/src/locale.js
+++ b/src/locale.js
@@ -106,8 +106,9 @@ export default class Locale {
 		 *		t( { string: 'image', context: 'Add/Remove image' } );
 		 *
 		 * @method #t
-		 * @param {String|module:utils/translation-service~Message} message A message that will be localized.
-		 * @param {Array.<String>} [values] Values that should be used to interpolate the string.
+		 * @param {String|module:utils/translation-service~Message} message A message that will be localized (translated).
+		 * @param {Array.<String>} [values] An array of values that will fill message placeholders.
+		 * For messages supporting plural forms the first value will determine the plural form.
 		 * @returns {String}
 		 */
 		this.t = ( message, values ) => this._t( message, values );
@@ -138,12 +139,12 @@ export default class Locale {
 	}
 
 	/**
-	 * Base for the {@link #t} method.
+	 * An unbound version of the {@link #t} method.
 	 *
+	 * @private
 	 * @param {String|module:utils/translation-service~Message} message
 	 * @param {Array.<String>} [values]
 	 * @returns {String}
-	 * @private
 	 */
 	_t( message, values = [] ) {
 		if ( typeof message === 'string' ) {
--- a/src/locale.js
+++ b/src/locale.js
@@ -105,12 +105,11 @@ export default class Locale {
 		 *		t( { string: 'Add a space', plural: 'Add %0 spaces' }, 5 ); // 'Dodaj 5 spacji' for the Polish language.
 		 *		t( { string: '%1 a space', plural: '%1 %0 spaces' }, [ 2, 'Add' ] ); // 'Dodaj 2 spacje' for the Polish language.
 		 *
-		 * The message can provide a context using the `context` property when the message ids created from message strings
-		 * are not unique. When the `context` property is set the message id will be constructed in
-		 * the following way: `${ message.string }_${ message.context }`. This context will be also used
-		 * by translators later as an additional context for the translated message.
+		 *  * The message should provide an id using the `id` property when the message strings are not unique and their
+		 * translations should be different.
 		 *
-		 *		t( { string: 'image', context: 'Add/Remove image' } );
+		 *		translate( 'en', { string: 'image', id: 'ADD_IMAGE' } );
+		 *		translate( 'en', { string: 'image', id: 'AN_IMAGE' } );
 		 *
 		 * @method #t
 		 * @param {String|module:utils/translation-service~Message} message A message that will be localized (translated).
--- a/src/locale.js
+++ b/src/locale.js
@@ -84,24 +84,24 @@ export default class Locale {
 		 *		const t = locale.t;
 		 *		t( 'Label' );
 		 *
-		 * The message can be either a string or an object implementing the {@link module:translation-service~Message} interface.
+		 * The message can be either a string or an object implementing the {@link module:utils/translation-service~Message} interface.
 		 *
-		 * A Message may contain placeholders (`%<index>`) for values that are passed as the second argument.
+		 * The message may contain placeholders (`%<index>`) for values that are passed as the second argument.
 		 * `<index>` is the index in the `values` array.
 		 *
 		 *		t( 'Created file "%0" in %1ms.', [ fileName, timeTaken ] );
 		 *
-		 * A Message can provide a plural form using the `plural` property and a value - that should be always the first element
-		 * of the `values` array based on which the plural form of the target language should be picked. That property value will
-		 * be used as a default plural translation when the translation for the target language will be missing.
+		 * The message supports plural forms. To specify a plural form, use the `plural` property. Single or plural form
+		 * will be chosen depending on the first value from the passed `values`. The value of this property is
+		 * used as a default plural translation when the translation for the target language is missing.
 		 *
 		 *		t( { string: 'Add a space', plural: 'Add %0 spaces' }, [ spaces ] );
 		 *		t( { string: '%1 a space', plural: '%1 %0 spaces' }, [ spaces, 'Add' ] );
 		 *
-		 * A Message can provide a context using the `context` property when the message string may be not enough unique
-		 * across all messages. When the `context` property is set the message id will be constructed in
+		 * The message can provide a context using the `context` property when the message ids created from message strings are not unique.
+		 * When the `context` property is set the message id will be constructed in
 		 * the following way: `${ message.string }_${ message.context }`. This context will be also used
-		 * by translators later as a context for the message that should be translated.
+		 * by translators later as an additional context for the translated message.
 		 *
 		 *		t( { string: 'image', context: 'Add/Remove image' } );
 		 *
--- a/src/locale.js
+++ b/src/locale.js
@@ -77,7 +77,7 @@ export default class Locale {
 
 		/**
 		 * Translates the given message to the {@link #uiLanguage}. This method is also available in
-		 * {@link module:core/editor/editor~Editor#t} and {@link module:ui/view~View#t}.
+		 * {@link module:core/editor/editor~Editor#t Editor} and {@link module:ui/view~View#t View}.
 		 *
 		 * This method's context is statically bound to the `Locale` instance and **always should be called as a function**:
 		 *
@@ -86,16 +86,16 @@ export default class Locale {
 		 *
 		 * The message can be either a string or an object implementing the {@link module:utils/translation-service~Message} interface.
 		 *
-		 * The message may contain placeholders (`%<index>`) for value(s) that are passed as the second argument.
+		 * The message may contain placeholders (`%<index>`) for value(s) that are passed as a `values` parameter.
 		 * For an array of values the `%<index>` will be changed to an element of that array at the given index.
 		 * For a single value passed as the second argument, only the `%0` placeholders will be changed to the provided value.
 		 *
 		 *		t( 'Created file "%0" in %1ms.', [ fileName, timeTaken ] );
 		 * 		t( 'Created file "%0", fileName );
 		 *
-		 * The message supports plural forms. To specify a plural form, use the `plural` property. Single or plural form
+         * The message supports plural forms. To specify the plural form, use the `plural` property. Singular or plural form
 		 * will be chosen depending on the first value from the passed `values`. The value of this property is used
-		 * as a default plural translation when the translation for the target language is missing. Therefor, it should be a number.
+		 * as a default plural translation when the translation for the target language is missing. Therefore, it should be a number.
 		 *
 		 *		t( { string: 'Add a space', plural: 'Add %0 spaces' }, 1 ); // 'Add a space' for the English language.
 		 *		t( { string: 'Add a space', plural: 'Add %0 spaces' }, 5 ); // 'Add 5 spaces' for the English language.
--- a/src/locale.js
+++ b/src/locale.js
@@ -91,10 +91,10 @@ export default class Locale {
 		 *		t( 'Label' );
 		 *
 		 * @method #t
-		 * @param {String} str The string to translate.
+		 * @param {String} message The string to translate.
 		 * @param {String[]} [values] Values that should be used to interpolate the string.
 		 */
-		this.t = ( ...args ) => this._t( ...args );
+		this.t = ( message, values ) => this._t( message, values );
 	}
 
 	/**
@@ -124,10 +124,19 @@ export default class Locale {
 	/**
 	 * Base for the {@link #t} method.
 	 *
+	 * @param {Object|String} message
+	 * @param {String[]} [values]
 	 * @private
 	 */
-	_t( str, values ) {
-		let translatedString = translate( this.uiLanguage, str );
+	_t( message, values = [] ) {
+		if ( typeof message === 'string' ) {
+			message = { string: message };
+		}
+
+		const hasPluralForm = !!message.plural;
+		const amount = hasPluralForm ? values[ 0 ] : 1;
+
+		let translatedString = translate( this.uiLanguage, message, amount );
 
 		if ( values ) {
 			translatedString = translatedString.replace( /%(\d+)/g, ( match, index ) => {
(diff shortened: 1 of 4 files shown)


src/locale.js
```diff
@@ -91,10 +91,10 @@ export default class Locale {
 		 *		t( 'Label' );
 		 *
 		 * @method #t
-		 * @param {String} str The string to translate.
+		 * @param {String} message The string to translate.
```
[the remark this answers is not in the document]
ma2ciek: Yes... There are still places with missing docs. I'll focus on them after reverting 100% CC.

tests/locale.js
```diff
@@ -115,49 +114,67 @@ describe( 'Locale', () => {
...
 
-		it( 'does not interpolate those placeholders for which values has not been passed', () => {
+		it( 'should interpolate a message with provided values', () => {
 			const t = locale.t;
 
+			expect( t( '%0 - %0', [ 'foo' ] ) ).to.equal( 'foo - foo' );
+			expect( t( '%1 - %0 - %2', [ 'a', 'b', 'c' ] ) ).to.equal( 'b - a - c' );
+
+			// Those test make sure that if %0 is really to be used, then it's going to work.
+			// It'd be a super rare case if one would need to use %0 and at the same time interpolate something.
```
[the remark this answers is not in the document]
ma2ciek: We could provide some escaping mechanism in the follow-up, I'm afraid that yours could be impossible though :smile: However with a character other than `\` which won't be lost during string creation should be possible.

src/translation-service.js
```diff
@@ -93,13 +133,26 @@ export function _clear() {
...
+		!!window.CKEDITOR_TRANSLATIONS[ language ].dictionary[ messageId ]
 	);
 }
 
 function getNumberOfLanguages() {
 	return Object.keys( window.CKEDITOR_TRANSLATIONS ).length;
 }
+
+/**
+ * The internalization message interface. A translation for the given language can be found.
```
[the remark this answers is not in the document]
ma2ciek: Ouch.

src/translation-service.js
```diff
@@ -53,18 +63,32 @@ export function add( language, translations ) {
...
  * @param {String} language Target language.
- * @param {String} translationKey String that will be translated.
+ * @param {module:utils/translation-service~Message|string} message A message that will be translated.
+ * @param {Number} [amount] A number of elements for which a plural form should be picked from the target language dictionary.
  * @returns {String} Translated sentence.
  */
-export function translate( language, translationKey ) {
+export function translate( language, message, amount = 1 ) {
+	if ( typeof message === 'string' ) {
+		// TODO
```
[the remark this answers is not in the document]
ma2ciek: > Shouldn't we simply expect that the function takes an object from now on? Why still let using strings?

Theoretically, someone could use the `translate()` function in the past, however now I think that this is mostly a protected method and we should mark it as protected and don't provide any deprecation warnings there.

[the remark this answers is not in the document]
ma2ciek: We use this function only in the `locale.t()` method, the separation was made to manipulate the global `window.CKEDITOR_TRANSLATIONS` object only in the `translation-service` file.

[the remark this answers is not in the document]
scofalik: So, who about we make a backward incompatible change and just keep it simple? I assume this wasn't a popular function :D.

tests/locale.js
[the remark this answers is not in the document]
scofalik: Okay let's skip that for now. We will add it if there will be a request for it.

src/locale.js
```diff
@@ -76,23 +76,40 @@ export default class Locale {
...
+		 * Translates the given message to the {@link #uiLanguage}. This method is also available in
 		 * {@link module:core/editor/editor~Editor#t} and {@link module:ui/view~View#t}.
 		 *
-		 * The strings may contain placeholders (`%<index>`) for values which are passed as the second argument.
+		 * This method's context is statically bound to the `Locale` instance and **always should be called as a function**:
+		 *
+		 *		const t = locale.t;
+		 *		t( 'Label' );
+		 *
+		 * The message can be either a string or an object implementing the {@link module:translation-service~Message} interface.
```
scofalik, line 87: I think this link may be wrong, isn't it module:utils/translation-service~Message?

src/locale.js
```diff
@@ -76,23 +76,40 @@ export default class Locale {
...
+		 *
+		 * A Message may contain placeholders (`%<index>`) for values that are passed as the second argument.
```
scofalik, line 89: A Message -> The message? (As in the previous sentence above.) (Same below.)

src/locale.js
```diff
@@ -76,23 +76,40 @@ export default class Locale {
...
+		 * A Message can provide a plural form using the `plural` property and a value - that should be always the first element
+		 * of the `values` array based on which the plural form of the target language should be picked. That property value will
+		 * be used as a default plural translation when the translation for the target language will be missing.
 		 *
-		 *		const t = this.t;
-		 *		t( 'Label' );
+		 *		t( { string: 'Add a space', plural: 'Add %0 spaces' }, [ spaces ] );
+		 *		t( { string: '%1 a space', plural: '%1 %0 spaces' }, [ spaces, 'Add' ] );
+		 *
+		 * A Message can provide a context using the `context` property when the message string may be not enough unique
```
scofalik, line 101: You can't be "enough unique". You either are, or not. So, keep it simple: "when the message string ids not unique".

src/locale.js
```diff
@@ -76,23 +76,40 @@ export default class Locale {
...
+		 *
+		 * A Message may contain placeholders (`%<index>`) for values that are passed as the second argument.
 		 * `<index>` is the index in the `values` array.
 		 *
-		 *		editor.t( 'Created file "%0" in %1ms.', [ fileName, timeTaken ] );
+		 *		t( 'Created file "%0" in %1ms.', [ fileName, timeTaken ] );
 		 *
-		 * This method's context is statically bound to Locale instance,
-		 * so it can be called as a function:
+		 * A Message can provide a plural form using the `plural` property and a value - that should be always the first element
```
scofalik, line 94:
```suggestion
		 * The message support plural forms. To specify a plural form, use `plural`. Single or plural form will be chosen depending on the first value from the passed `values`.
```

scofalik: How about that?

scofalik: (PS. Introduce the change in code, not by committing this suggestion, as the length of this line will probably be too big.)

scofalik: (And merge my proposal with the lines below, where you describe how it may work for different languages ;).)

src/locale.js
```diff
@@ -76,23 +76,40 @@ export default class Locale {
...
 		 *
-		 *		const t = this.t;
-		 *		t( 'Label' );
+		 *		t( { string: 'Add a space', plural: 'Add %0 spaces' }, [ spaces ] );
+		 *		t( { string: '%1 a space', plural: '%1 %0 spaces' }, [ spaces, 'Add' ] );
+		 *
+		 * A Message can provide a context using the `context` property when the message string may be not enough unique
+		 * across all messages. When the `context` property is set the message id will be constructed in
+		 * the following way: `${ message.string }_${ message.context }`. This context will be also used
+		 * by translators later as a context for the message that should be translated.
```
scofalik, line 104: "as an additional context for the translated message".

src/translation-service.js
```diff
@@ -93,13 +133,26 @@ export function _clear() {
...
 }
 
 function getNumberOfLanguages() {
 	return Object.keys( window.CKEDITOR_TRANSLATIONS ).length;
 }
+
+/**
+ * The internalization message interface. A translation for the given language can be found.
+ *
+ * TODO
```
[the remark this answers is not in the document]
ma2ciek: I removed the TODO there.

src/translation-service.js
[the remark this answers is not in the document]
ma2ciek: Fixed.

src/translation-service.js
```diff
@@ -17,26 +17,32 @@ if ( !window.CKEDITOR_TRANSLATIONS ) {
 }
 
 /**
- * Adds translations to existing ones or overrides the existing translations.
+ * Adds translations to existing ones or overrides the existing translations. these translations will later
```
scofalik, line 20:
```suggestion
 * Adds translations to existing ones or overrides the existing translations. These translations will later
```

src/translation-service.js
```diff
@@ -17,26 +17,32 @@ if ( !window.CKEDITOR_TRANSLATIONS ) {
...
+ * be available for the {@link module:utils/locale~Locale#t `t()`} function.
  *
- * These translations will later be available for the {@link module:utils/locale~Locale#t `t()`} function.
+ * The `translations` is an object which consists of a `message id - translation` pairs. Note that message id can be
```
scofalik, line 23:
```suggestion
 * The `translations` is an object which consists of a `messageId: translation` pairs. Note that message id can be
```

src/translation-service.js
```diff
@@ -17,26 +17,32 @@ if ( !window.CKEDITOR_TRANSLATIONS ) {
...
+ * either constructed either from the message string or from the message string and message context in the following form:
```
scofalik, line 24:
```suggestion
 * either constructed either from the message string or from the message string and the message context in the following form:
```

src/translation-service.js
```diff
@@ -17,26 +17,32 @@ if ( !window.CKEDITOR_TRANSLATIONS ) {
...
+ * `<messageString>_<messageContext>` (this happens rarely and mostly for short messages or messages with placeholders).
+ * Since the editor displays only the message string the message context can be found either in the source code or in the
```
scofalik, line 26:
```suggestion
 * Since the editor displays only the message string, the message context can be found either in the source code or in the
```

src/translation-service.js
```diff
@@ -17,26 +17,32 @@ if ( !window.CKEDITOR_TRANSLATIONS ) {
...
  *
- * You should also specify the third argument (the `getPluralForm` function) that will be used to determine the plural form if the
- * language is not supported by CKEditor 5 yet.
+ * You should also specify the third argument (the `getPluralForm` function) that will be used to determine the plural form if no
+ * language file was loaded for that language. All language files coming from CKEditor 5 sources will have this option set, so
+ * these plural form rules will be reused by other translations added to the registered languages.
  *
  * 		add( 'pl', {
- *	 		'Add space': [ 'Dodaj spację', 'Dodaj %0 spacje', 'Dodaj %0 spacji' ]
+ *	 		// Translations.
```
scofalik, line 45:
```suggestion
 *	 		// ... Translations.
```

ma2ciek:
I used the `amount` word across these two files. But maybe the `quantity` fits better?

cc @scofalik

src/translation-service.js
```diff
@@ -9,62 +9,127 @@
...
+ *	 		'Add space': [ 'Dodaj spację', 'Dodaj %0 spacje', 'Dodaj %0 spacji' ]
+ * 		} );
+ *
+ * You should also specify the third argument (the `getPluralForm` function) that will be used to determine the plural form if no
+ * language file was loaded for that language. All language files coming from CKEditor 5 sources will have this option set, so
+ * these plural form rules will be reused by other translations added to the registered languages.
+ *
+ * 		add( 'pl', {
+ *	 		// ... Translations.
+ * 		}, n => n == 1 ? 0 : n % 10 >= 2 && n % 10 <= 4 && ( n % 100 < 10 || n % 100 >= 20 ) ? 1 : 2 );
```
mlewand, line 46: This is some serious one liner action going right here :smiley: Especially when nesting multiple ternary operators.

Instead can we start with english language example first? this will be simple enough, set the ground for this difficult one like polish. I wouldn't mind asking people for other language which has a simpler rules - because this will be a madness for non-polish speaker.

src/translation-service.js
```diff
@@ -9,62 +9,127 @@
...
- *			'OK': 'OK',
- *			'Cancel [context: reject]': 'Anuluj'
+ *			'Cancel': 'Anuluj',
+ *			'image_Insert image': 'obraz', // Note that the `Insert image` comes from the message context.
  *		} );
  *
+ * If the message is supposed to support various plural forms, make sure to provide an array with the single form and all plural forms:
+ *
+ *		add( 'pl', {
+ *	 		'Add space': [ 'Dodaj spację', 'Dodaj %0 spacje', 'Dodaj %0 spacji' ]
```
mlewand, line 37: What is the equivalent of adding it using our `t()` function?

In the docs I've seen only like:

```plaintext
translate( 'en', { string: 'Add a space', plural: 'Add %0 spaces' }, 3 );
```

Which seems to be limited to one plural form.

src/locale.js
```diff
@@ -76,25 +76,49 @@ export default class Locale {
...
+		 * the following way: `${ message.string }_${ message.context }`. This context will be also used
+		 * by translators later as an additional context for the translated message.
+		 *
+		 *		t( { string: 'image', context: 'Add/Remove image' } );
 		 *
 		 * @method #t
-		 * @param {String} str The string to translate.
-		 * @param {String[]} [values] Values that should be used to interpolate the string.
+		 * @param {String|module:utils/translation-service~Message} message A message that will be localized (translated).
+		 * @param {String|Number|Array.<String|Number>} [values] A value or an array of values that will fill message placeholders.
```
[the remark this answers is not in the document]
ma2ciek: argue with @scofalik :smile: 

For me the cost of writing `[]` around one argument is rather low. But it's true that in our use cases it's ~90% of all cases.

src/locale.js
```diff
@@ -76,25 +76,49 @@ export default class Locale {
...
-		 *		const t = this.t;
-		 *		t( 'Label' );
+		 * The message may contain placeholders (`%<index>`) for value(s) that are passed as the second argument.
+		 * For an array of values the `%<index>` will be changed to an element of that array at the given index.
+		 * For a single value passed as the second argument, only the `%0` placeholders will be changed to the provided value.
+		 *
+		 *		t( 'Created file "%0" in %1ms.', [ fileName, timeTaken ] );
+		 * 		t( 'Created file "%0", fileName );
+		 *
+		 * The message supports plural forms. To specify a plural form, use the `plural` property. Single or plural form
```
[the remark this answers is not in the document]
ma2ciek: Yep. You're right.

ma2ciek:
>  Interesting to see that you decided to handle such a tricky case as multiple plural version - what is use case for it? 🙂

There's a lot of use cases in our track changes code where the `add spaces (3)` doesn't look as good as we want.

>  - this change introduces a breaking change to public API that should be noted.

You mean the whole format of the `window.CKEDITOR_TRANSLATIONS`? Certainly it needs to be mentioned.

> why did you go with the `string` + `context` prop combo rather than [proposed `id` property](https://github.com/ckeditor/ckeditor5/issues/6526#issuecomment-608406382) in the end?
> 
> *   I'm a little tentative about using `context` + `string` to build a translation id, as context seems to be prone to changes during product lifetime (like, stupid "a", "an", "the" changes). Instead I think that id will be a better option, and since we'll have to inline all contextes, why not inlining all the ids in a single run?

The `context` + `string` combo will be an id only in situations where the `context` will be a part of the `t( messageObject )` form. Therefore it will be a little bit shorter than our long contexts and hopefully it won't be changed so often. But that's for sure a thing to remember.

And why it's no `id`? After some f2f talks with @scofalik we decided to change it as the `context` + `string` should be unique enough for creating a message id, and this doesn't require writing the same twice (as the id and the context would probably quite similar). But there're some obvious cons of it as it's less verbose, creates context + message, that's hard to change in the future and maybe not intuitive. As I think now that changing the message string or message context will break translations I'm leaning towards the id :joy: 

> This is a major change and getting it from the API docs alone is rather difficult task. We should add a guide/doc on i11n matter, which would include feature added with this PR. It's a perfect moment to add this guide in this release (ofc should be a followup task).

I was thinking about it too. There're a few things worth noting. We have an internal guide (which is going to be outdated after this task). There's a guide for setting the UI language - https://ckeditor.com/docs/ckeditor5/latest/features/ui-language.html, but we lack a guide for a 3rd-party plugin creator.

src/translation-service.js
ma2ciek: The plural form in a translation message specifies that the message supports plural version and specifies the default plural form for missing translations. It doesn't specifies the number of plural forms.

So the `translate( 'pl', { string: 'Add a space', plural: 'Add %0 spaces' }, 3 );` should return `Dodaj 3 spacje` in your example.

scofalik:
As far as id is concerned... Isn't it a problem right now as well? You introduce some translation as `t( 'Foobar' )` and then it becomes a problem if we would like to change it. Isn't it the same here? How it was solved until now?

What I don't like about id is that it is repeating the same information again :S.

ma2ciek:
> As far as id is concerned... Isn't it a problem right now as well? You introduce some translation as `t( 'Foobar' )` and then it becomes a problem if we would like to change it. Isn't it the same here? How it was solved until now?

Yes, but only for message strings as message contexts (longer and therefore changing more often) are kept in `contexts.json` file. Note that in our cases `context` and `string` are long strings  and are changing 3 times a day currently :smile: 

Actually having the `messageId` in track changes would allow us to change the message string or message context without any trouble in the future. With the `context` + `string` where any one can't change the situation is quite opposite. When a string would be changes without changing id Transifex would also remove all translations but with better effect - https://docs.transifex.com/projects/updating-content - it would preserve the history of string translation and so on.

ma2ciek:
That's why I chose this new interface for the `t()` function - you can still extend the message with the `id` property in the future :smile:

scofalik:
If you thing it is correct, go with it.

src/locale.js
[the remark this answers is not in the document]
Reinmar: I can see we didn't support this shorter form previously. But I agree that in great majority of cases there will be just one placeholder, so this may be a useful addition. OTOH, this function is used so few times that I'd personally do KISS, as @mlewand mentions.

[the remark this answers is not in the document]
mlewand: After some extra f2f talking we decided we're ok with the way it was proposed by @ma2ciek in this PR (so both primitive and array variants are accepted).

src/translation-service.js
mlewand: OK, so to ensure I get this right: `t()` can pass only one plural form (because English supports only one plural form) thus multiple plural forms will be _always_ added using the [`translation-service#add()`](https://ckeditor.com/docs/ckeditor5/18.0.0/api/module_utils_translation-service.html#static-function-add) method, right?

If that's the case, this conversation is resolved.

mlewand:
> There's a lot of use cases in our track changes code where the `add spaces (3)` doesn't look as good as we want.

:+1: 

> You mean the whole format of the `window.CKEDITOR_TRANSLATIONS`? Certainly it needs to be mentioned.

That's right, change of structure in this global variable is a BC. So needs to be reflected in changelog.

> This is a major change and getting it from the API docs alone is rather difficult task. We should add a guide/doc on i11n matter

Guide task extracted to ckeditor/ckeditor5#6646 - this is something we can get back to later on.

src/locale.js
```diff
@@ -122,23 +146,38 @@ export default class Locale {
...
-			translatedString = translatedString.replace( /%(\d+)/g, ( match, index ) => {
-				return ( index < values.length ) ? values[ index ] : match;
-			} );
+		if ( typeof message === 'string' ) {
+			message = { string: message };
 		}
 
-		return translatedString;
+		const hasPluralForm = !!message.plural;
+		const quantity = hasPluralForm ? values[ 0 ] : 1;
```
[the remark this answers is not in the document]
ma2ciek: For me as a consumer you either want to pass a quantity as a number (plus interpolation) or a string that will be a part of the interpolation.

src/translation-service.js
ma2ciek: Right!

src/translation-service.js
```diff
@@ -9,62 +9,130 @@
...
 }
 
 /**
- * Adds translations to existing ones.
- * These translations will later be available for the {@link module:utils/translation-service~translate `translate()`} function.
+ * Adds translations to existing ones or overrides the existing translations. These translations will later
+ * be available for the {@link module:utils/locale~Locale#t `t()`} function.
+ *
+ * The `translations` is an object which consists of a `messageId: translation` pairs. Note that the message id can be
+ * either constructed either from the message string or from the message id if it was passed
```
mlewand, line 24:
```suggestion
 * either constructed from the message string or from the message id if it was passed
```

